PR #4, opened 2020-07-17
gcp collector .tar generation

JingboWangGoogle:
This PR provides the functionality to generate a tarball file containing the collector binary and a cloud monitoring compatible config file. This generation process is performed inside a docker container with one make command


linux/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN apt-get -y update \
+    && apt-get -y install \
+        gettext-base \
+    && apt-get -y clean \
+    && rm -rf /var/lib/apt/lists/*
+
+ADD "generate_tar.sh" "."
+ADD "gcp-config.yaml" "."
+
+CMD ["./generate_tar.sh"]
```
james-bebbington, line 12: Usually the sepration I've seen is:

- Dockerfile includes all dependencies needed for building
- Build scripts are mounted and run inside the docker container

The dependencies change infrequently so you rarely need to rebuild your Dockerfile, whereas your build scripts are likely to change slightly more often.

That doesn't necessarily make as big of a difference as you would think though as each command in a Docker file gets cached, so if you only change the build scripts, only the most recent commands (3 in this Dockerfile) would need to be recreated.

But ideally you don't even want to have to rebuild your Dockerfile on each build.

Makefile
```diff
@@ -8,3 +8,15 @@ DOCKER_RUN=docker run -v $(CURDIR):/mnt otelopscol-build
...
 package-googet:
 	$(DOCKER_RUN) /bin/bash -c "cd /mnt; make -f .build/Makefile VERSION=$(VERSION) ARCH=$(ARCH) package-googet"
+
+# gcp collector related variables
+IMAGE_NAME=gcpcol_image
+CONTAINER_NAME=gcpcol_container
+
+.PHONY: gcpcol
+gcpcol:
+	docker build -t $(IMAGE_NAME) ./linux
```
james-bebbington, line 18: As per the comment above, while this works, you generally don't want to rebuild your Dockerfile on each build, so would be good to separate this line into a diff make command.

Makefile
```diff
@@ -8,3 +8,15 @@ DOCKER_RUN=docker run -v $(CURDIR):/mnt otelopscol-build
...
+	docker run -i --name=$(CONTAINER_NAME) $(IMAGE_NAME)
+	docker cp $(CONTAINER_NAME):/go/gcp-otel.tar.gz .
```
james-bebbington, line 20: Interesting, I guess this works just as well given your setup, but I would normally mount your local directory as a volume in the docker container instead.

(see the `-v $(CURDIR):/mnt` flag above)

That way you can run local scripts in the docker container and any files generated in the container (inside the mounted volume) will exist on your local system as well.

Makefile
```diff
@@ -8,3 +8,15 @@ DOCKER_RUN=docker run -v $(CURDIR):/mnt otelopscol-build
...
+	docker rmi -f $(IMAGE_NAME)
```
james-bebbington, line 21: Any reason to remove the image?

linux/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.14-stretch
+
+RUN apt-get -y update \
+    && apt-get -y install \
+        gettext-base \
+    && apt-get -y clean \
+    && rm -rf /var/lib/apt/lists/*
```
james-bebbington, line 7: FYI I had just added this to the other dockerfile to use `envsubst`

linux/Dockerfile
james-bebbington: Since Go can be cross-compiled for any OS, I'm keen to keep this a bit simpler and use the existing Dockerfile in this repo for builds on all environments unless we find there's significantly different dependencies

linux/generate_tar.sh
```diff
@@ -0,0 +1,32 @@
...
+# locate config file
+CONFIG_FILE="gcp-config.yaml"
+if [ ! -e "$CONFIG_FILE" ]
+then 
+    echo "missing required config file: $CONFIG_FILE"
+    exit 1
+fi
+
+# get the repo into local
+git clone https://github.com/JingboWangGoogle/opentelemetry-collector-contrib
```
james-bebbington, line 12: Why not just build the code in this repo?

linux/gcp-config.yaml
```diff
@@ -0,0 +1,161 @@
...
+    scrapers:
+      cpu:
+        report_per_cpu: true
+      load:
+      memory:
+      disk:
+      filesystem:
+      network:
+      process:
+  # add receivers from Nic to extend the ability for MySQL, JVM and Apache
```
james-bebbington, line 14: In that doc I think I implied we'd have a separate config file for each packaging method, but I've since realized that's probably not necessary. We just need a separate config for each OS I think (i.e. one for Linux & one for Windows).

That makes me think:

- Remove Dockerfile and just use existing one
- Move `generate_tar.sh` into a "tar" directory
- Create a config directory, put this in there, and rename to maybe "config-gce-linux.yaml"?

james-bebbington: Also note this config file is quite outdated now - see https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/2

Makefile
JingboWangGoogle: I have updated the make file to reuse the build command, and have the gcpcol command just running the image.

linux/Dockerfile
JingboWangGoogle: So based on my understanding, we want this dockerfile to only contain the setups of the environment. In this case, how should I go about it if I want this container to
1. pull down a repo
2. build the binary
3. compress files and folders into a tar file

JingboWangGoogle: currently reusing the dockerfile in .build. mount and execute the script when running the container.

Makefile
JingboWangGoogle: perfect

Makefile
JingboWangGoogle: image not removed anymore since it doesn't need to be rebuilt everytime.

linux/Dockerfile
JingboWangGoogle: I don't have a dockerfile anymore

linux/generate_tar.sh
JingboWangGoogle: Changed to, instead of cloning the repo, directly build the binary here.

JingboWangGoogle:
@draffensperger @Nicolas-MacBeth .tar file was mistakenly committed here, and it is now removed. The .ignore file is also updated to ignore the .tar file

tar/generate_tar.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# locate config file
+CONFIG_FILE="config-example.yaml"
```
james-bebbington, line 4: I would make `CONFIG_FILE` a parameter with "config-example.yaml" as the default value, so users can pass in a custom config if they like.

tar/generate_tar.sh
```diff
@@ -0,0 +1,20 @@
...
+    echo "missing required config file: $CONFIG_FILE"
+    exit 1
+fi
+
+# move the binary back to the root directory
+cp config/$CONFIG_FILE tar/
+cp bin/$OTELCOL_BINARY tar/
+
+# compress the binary and the config into a .tar file
+cd tar && tar -cvzf gcp-otel.tar.gz $OTELCOL_BINARY $CONFIG_FILE
```
james-bebbington, line 17: For consistency, can you move the tarball into the `dist` folder instead

.gitignore
```diff
@@ -25,3 +25,6 @@ coverage.html
 # Wix
 *.wixobj
 *.wixpdb
+
+# Tarball file
+*.tar.gz
```
james-bebbington, line 30: See other comment about moving the generated tarball into the `dist` folder in which case you wouldn't need this as the `dist` directory is already ignored above

Makefile
```diff
@@ -48,16 +54,18 @@ pack-googet:
...
+	./tar/generate_tar.sh
 
 # --------------------
 #  Create build image
 # --------------------
 
 .PHONY: docker-build-image
-build-image:
-	docker build -t otelopscol-build ./.build
+docker-build-image:
```
james-bebbington, line 67: Thanks for the fix 😄

Makefile
```diff
@@ -48,16 +54,18 @@ pack-googet:
 	goopack -output_dir ./dist <(envsubst < ./.build/googet/google-cloudops-opentelemetry-collector.goospec)
 
 # tarball (Linux)
-
-# TODO
+.PHONY: linux-tarball
+linux-tarball:
+	make build
```
james-bebbington, line 59: I think it's convenient to separate the build and packaging commands. To be consistent with above can you make it:

```make
.PHONY: build-tarball
build-tarball:
	build package-tarball

.PHONY: package-tarball
package-tarball:
	./tar/generate_tar.sh
```
Also I think this works on OSs other than just Linux right? So probably drop the "(Linux)" comment above.

And also pls update the googet target names above so they are consistent (`build` & `package` instead of `package` & `pack`)

james-bebbington:
> Will the `gcp-config.yaml` be for both Linux and Windows? Do we need separate configs for each OS? James Bebbington And can we name them based on the OS?

I think we will need diff configs per OS when we add default application integrations, etc. I don't think we would need anything more fine grained than that?

tar/generate_tar.sh
JingboWangGoogle: moved CONFIG_FILE declaration to make, so the variable can be passed in through make command

tar/generate_tar.sh
JingboWangGoogle: tarball is now generated in the dist folder

Makefile
JingboWangGoogle: updated. One remaining question thought. The syntax `build package-tarball` doesn't seem to work. I had to change this to `make build``make package-tarball` to have it working. Am I doing something wrong?

james-bebbington: Yea sorry makefiles are very sensitive. You can probably do:

```make
build-tarball: build package-tarball
```
or
```make
build-tarball:
	make build package-tarball
```
but not
```make
build-tarball:
	build package-tarball
```

james-bebbington: I was probably guilty of not testing that googet make command I checked in before 😮

james-bebbington:
@JingboWangGoogle I will merge this to unblock other work but can you pls move the `tar` folder into `.build/tar` in a follow-up PR

james-bebbington:
Done here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/7

config/config-example.yaml
```diff
@@ -0,0 +1,39 @@
...
+          - action: aggregate_label_values
+            label: state
+            aggregated_values: [slab_reclaimable, slab_unreclaimable]
+            new_value: slab
+            aggregation_type: sum
+
+exporters:
+  # export to Cloud Monitoring using the agent prefix
+  stackdriver:
+    project: "otel-gce-metrics" # change this to the project name of the Monitoring project
```
draffensperger, line 27: Late drive-by-comment here, but we should make it so that running on GCP allows this project to be auto-detected based on the authentication credential. Having to go in and set it sort of a rough edge in the process. I think it would be worth contributing that change if needed.

PR #6, opened 2020-07-22
Initial configuration files for 3 main applications and exporters build script

Nicolas-MacBeth:
OpenTelemetry Collector config for JVM, MySQL, Apache, DogStatsD, host and custom metrics collection.

Update: Waiting on collector-contrib version upgrade to see if it fixes build error before merging.


.build/tar/generate_tar.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-CONFIG_FILE=${CONFIG_FILE:-config-example.yaml}
+CONFIG_FILE=${CONFIG_FILE:-config-mysql_apache_jvm.yaml}
```
james-bebbington, line 3: Should we change the default config file to this?

Instead users can select which config to use when running the make command:
```bash
make CONFIG_FILE=config-mysql_apache_jvm.yaml build-tarball
```

Makefile
```diff
@@ -28,8 +28,7 @@ ifeq ($(BUILD_IMAGE_NAME),)
 BUILD_IMAGE_NAME=otelopscol-build
 endif
 
-OTELCOL_BINARY=google-cloudops-opentelemetry-collector
-
+OTELCOL_BINARY=google-cloudops-opentelemetry-collector_$(GOOS)_$(GOARCH)$(EXTENSION)
```
james-bebbington, line 31: I'm not super familiar with make or bash, but for the case where:

- You are on Linux (GOOS defaults to linux)
- You run build-goo (which overrides GOOS to windows)

Does this build a binary called `google-cloudops-opentelemetry-collector_windows_...` or `google-cloudops-opentelemetry-collector_linux_...`?

(I'm not sure if the variable is set once initially or if this is re-run for each target that's called)

Can you check? If this works, then it's okay to leave as you have it

.gitignore
```diff
@@ -1,5 +1,6 @@
 bin/
 dist/
+prometheus_exporters/
```
james-bebbington, line 3: I don't think this is needed? You only ever create that directory inside `dist`

Makefile
JingboWangGoogle: addressed here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/9#discussion_r464748585

.build/tar/generate_tar.sh
JingboWangGoogle: I thought with the current statement, the user can still specify which config to use. What we have here is just the default one in case the user doesn't specify a config file while running the make command?

Nicolas-MacBeth: Just as Jingbo said, the user can still pass which config file they want, we're just here changing the default!

james-bebbington: Sorry my comment wasn't very clear. I was just asking why are we changing the default? Wouldn't the other one make more sense as the default? (relevant to more users)

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,83 @@
+# This config hopes to mimick the Cloud Monitoring Agent by collecting metrics from popular third-party services the Agent already supports
```
james-bebbington, line 1: `hopes` is not a very optimistic word 😄

james-bebbington: Should this config file include everything in Jingbo's config file as well?

Nicolas-MacBeth: Yeah, we'll merge them into one!

.build/tar/generate_tar.sh
Nicolas-MacBeth: Yes, we'll revert to the other config as default.

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
...
+cp mysqld_exporter/mysqld_exporter dist/prometheus_exporters/mysqld_exporter
+rm -rf mysqld_exporter
+
+# Clone Apache exporter, build the binary and put it in the folder
+echo "Cloning and building the Apache exporter"
+
+git clone https://github.com/Lusitaniae/apache_exporter.git -q
+cd apache_exporter
+make >/dev/null
+cd ..
```
james-bebbington, line 45: Consider
```suggestion
make -C apache_exporter
```

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "Building Prometheus exporters"
+
+# Clone mysqld exporter, build the binary and put it in the folder
+echo "Cloning and building the MySQL exporter"
+
+git clone https://github.com/prometheus/mysqld_exporter.git -q
+cd mysqld_exporter
+make >/dev/null
+cd ..
```
james-bebbington, line 35: Consider:
```suggestion
make -C apache_exporter
```

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Not found: dist folder, creating the folder dist"
+    mkdir dist
+fi
+
+# Check for prometheus_exporters folder
+if [ ! -d "dist/prometheus_exporters" ]
+then
+    echo "Not found: prometheus_exporters folder, creating the folder"
+    mkdir dist/prometheus_exporters
+fi
```
james-bebbington, line 25:
```suggestion
mkdir -p dist/prometheus_exporters
```

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# The use of >/dev/null is to supress stdout, but allow stderr to come thorugh
```
james-bebbington, line 3:
```suggestion
# The use of >/dev/null is to supress stdout, but allow stderr
```

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
...
+# This is because installing the tools and building makes a lot of output
```
james-bebbington, line 4: Is that a huge problem though? It might not be that bad to just print it all out?

Btw, how long does this script take to run? It is doing quite a lot

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "Installing tools needed"
+
+# Install git and maven for building the exporters
+sudo apt-get update >/dev/null
+sudo apt-get install -y git >/dev/null
+sudo apt-get install -y maven >/dev/null
```
james-bebbington, line 11: This in particular seems a weird thing to supress

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,60 @@
...
+rm -rf apache_exporter
+
+# Clone JVM exporter, build the binary and put it in the folder
+echo "Cloning and building the JVM exporter"
+
+git clone https://github.com/prometheus/jmx_exporter.git -q
+cd jmx_exporter
+version=$(sed -n -e 's#.*<version>\(.*-SNAPSHOT\)</version>#\1#p' pom.xml)
+mvn package >/dev/null
+cd ..
```
james-bebbington, line 56: Can you comment what `version= ...` does?

Also consider something like:
```suggestion
version=$(sed -n -e 's#.*<version>\(.*-SNAPSHOT\)</version>#\1#p' jmx_exporter/pom.xml)
mvn -f jmx_exporter/pom.xml package
```

Makefile
```diff
@@ -57,14 +56,19 @@ package-goo:
...
 # tarball
 
 .PHONY: build-tarball
-build-tarball: build package-tarball
+build-tarball: build build-exporters package-tarball
 
 .PHONY: package-tarball
 package-tarball:
-	./.build/tar/generate_tar.sh
+	bash ./.build/tar/generate_tar.sh
```
james-bebbington, line 71: Why is that change needed?

.build/tar/generate_tar.sh
```diff
@@ -11,22 +13,18 @@ then
     exit 1
 fi
 
-# check dist folder
-if [ ! -d "dist" ]
-then
-    echo "Not found: dist folder, creating the folder dist"
-    mkdir dist
-fi
```
james-bebbington, line 19: Isn't that still needed? (could be `mkdir -p dist`)

What if `package-tarball` is run directly?

Makefile
```diff
@@ -57,14 +56,19 @@ package-goo:
...
+# exporters
+.PHONY: build-exporters
+build-exporters:
+	bash ./.build/tar/build_exporters.sh
+
 # tarball
 
 .PHONY: build-tarball
-build-tarball: build package-tarball
+build-tarball: build build-exporters package-tarball
```
james-bebbington, line 67: Should `build-exporters` be optional? We need to provide a way to build a tarball without that don't we?

This is not necessarily required, but I think it would be even better if users could specify which exporters that want to include somehow (comma separated list?)

Another thought: shouldn't we clean `dist` before creating our tarball as there could be junk in there left over from other builds?

config/config-mysql_apache_jvm.yaml
james-bebbington: I would definitely change the wording to not say `hopes`

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,83 @@
...
+# and formatting the metrics in the same way, so they appear similarly in Cloud Monitoring. The services targeted here are MySQL, Apache and JVM.
+# Please remove/comment-out the prometheus_exec configs that you don't need in the receivers section as well as the list in the service/pipelines/metrics/receivers section.
+# Documentation specific to each exporter is found in the comment directly above its configuration
+
+receivers:
+
+  # https://github.com/prometheus/mysqld_exporter
+  prometheus_exec/mysql:
+    exec: ./prometheus_exporters/mysqld_exporter --web.listen-address=":{{port}}" --no-collect.global_variables  # this last flag reduces the amount of generated metrics, can be removed to enable all (600+) MySQL metrics
```
james-bebbington, line 10: I'd move that comment to the line above (likewise elsewhere in this config file where you have long trailing comments)

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,83 @@
...
+    exec: java -jar ./prometheus_exporters/jmx_exporter.jar {{port}} exporter_configs/jvm_config.yaml # see jvm_config.yaml for JMX configuration, notably the "hostPort" key
+    port: 9404
+
+processors:
+  filter:
+    metrics:
+      include:
+        match_type: regexp
+        metric_names: # Allowing all JVM and Apache metrics, but whitelisting MySQL ones to filter hundreds of unused ones
+                - mysql_global_status_commands_total
```
james-bebbington, line 32: Seems to be about 6 extra unnecessary spaces here.

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,83 @@
...
+exporters:
+  stackdriver:  
+    project:  # stackdriver will use the default GCE credentials
+    #metric_prefix: agent.googleapis.com/
+
+service:
+  pipelines:
+    metrics:
+      receivers: [prometheus_exec/mysql, prometheus_exec/apache, prometheus_exec/jvm]  # Remember to remove unwanted receivers from this list
+      processors: [filter, metricstransform/mysql, metricstransform/java]
```
james-bebbington, line 82: Missing `apache`?

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,83 @@
...
+      include:
+        match_type: regexp
+        metric_names: # Allowing all JVM and Apache metrics, but whitelisting MySQL ones to filter hundreds of unused ones
+                - mysql_global_status_commands_total
+                - mysql_global_status_threads_running
+                - mysql_global_status_handlers_total
+                - apache_.*
+                - jmx_.*
+                - java_.*
+                - promhttp_.*
```
james-bebbington, line 38: See pipeline suggestion below which I think would mean you can remove all these .* lines

config/config-mysql_apache_jvm.yaml
james-bebbington: Also consider making this three diff metrics pipelines, i.e.

```yaml
metrics/mysql:
      receivers: [prometheus_exec/mysql]
      processors: [filter/mysql, metricstransform/mysql]
      exporters: [stackdriver]

metrics/apache:
      receivers: [prometheus_exec/apache]
      processors: [metricstransform/apache]
      exporters: [stackdriver]

...
```

That will be slightly more efficient

Makefile
JingboWangGoogle: This was something I wanted to change also. When we ran this, especially in the docker one, I always get the permission denied error message. Even if I do a chmod -x command, I still get a permission denied, but if I add bash here, it runs successfully for me. Is there a convention for such things?

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: Yes! forsure, meant to do it last commit

.build/tar/build_exporters.sh
Nicolas-MacBeth: It is a lot of output, and lots of the specific exporter build scripts are really verbose and run tests suites on the build and such, which looks like a lot for not that much happening.

takes like 30-40 seconds

.build/tar/build_exporters.sh
Nicolas-MacBeth: I'll let that come through

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: If it is more efficient I will do this, but as of now it's meant for a user to be able to just remove a receiver from the list to modify the config to their needs, and having them in separate pipelines is a worst experience! If the performance gain is minimal I think this may be better for the config modification experience for the user!

Makefile
Nicolas-MacBeth: Yes, the exporters should be configurable, but given the lack of time we agreed that letting the user choose which receivers they don't want simply by removing them in the config is a viable option for now! The exporters would just be useless/not used in those cases.

I will make it optional though and good point on cleaning the dist/ folder

config/config-mysql_apache_jvm.yaml
james-bebbington: Why would that be a worse experience? Instead of deleting a receiver they could just delete a pipeline?

I feel like it would make the config file better modularised / readable which should be a better user experience.

Makefile
james-bebbington: Not sure, but that seems fine for now

.build/tar/build_exporters.sh
james-bebbington: Oh cool that's a lot faster than I expected

Makefile
james-bebbington: If users aren't going to use an exporter they wouldn't want to bundle it in the tar file at all. I guess its not a huge issue, but I would greatly prefer it if we at least had a make command for with exporters and without, which should be relatively trivial I think?

james-bebbington: Nvm just saw you already did that

.build/tar/build_exporters.sh
Nicolas-MacBeth: Was going off of memory, just re-ran it and took a little less than a minute!

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: Oh I was thinking they'd ignore un-needed receivers by simply removing them from the same list of `receivers` at the bottom, but I'll modify it for modularity!

james-bebbington: To me removing a pipeline is just as straightforward as removing the receiver.

Seems like you only added the `mysql` pipeline atm?

.build/tar/build_exporters.sh
```diff
@@ -0,0 +1,53 @@
...
+version=$(sed -n -e 's#.*<version>\(.*-SNAPSHOT\)</version>#\1#p' jmx_exporter/pom.xml)
+mvn -f jmx_exporter/pom.xml package >/dev/null
+cp jmx_exporter/jmx_prometheus_httpserver/target/jmx_prometheus_httpserver-${version}-jar-with-dependencies.jar dist/prometheus_exporters/jmx_exporter.jar
+rm -rf jmx_exporter
+
+# StatsD exporter
+echo "Cloning and building the StatsD exporter"
+
+git clone https://github.com/prometheus/statsd_exporter.git -q
+cd statsd_exporter/
```
james-bebbington, line 47: Can you make it consistent with the others? i.e. `make -C statsd_exporter build >/dev/null`

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,93 @@
...
+  # https://github.com/prometheus/jmx_exporter
+  prometheus_exec/jvm:
+    # see jvm_config.yaml for JMX configuration, notably the "hostPort" key
+    exec: java -jar ./prometheus_exporters/jmx_exporter.jar {{port}} exporter_configs/jvm_config.yaml
+    port: 9404
+
+  # https://github.com/prometheus/statsd_exporter
+  prometheus_exec/statsd:
+    exec: ./statsd_exporter
+    port: 9102
```
james-bebbington, line 32: You also need to add this to the pipelines (pending https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/6#discussion_r467769988)

.build/tar/build_exporters.sh
Nicolas-MacBeth: yeah, done! was merging Jingbo's changes into mine

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,307 @@
...
+      receivers: [prometheus_exec/jvm]
+      processors: [resourcedetection, metricstransform/jvm]
+      exporters: [stackdriver]
+
+    metrics/apache:
+      receivers: [prometheus_exec/apache]
+      processors: [resourcedetection, metricstransform/apache]
+      exporters: [stackdriver]
+
+    metrics/host:
```
james-bebbington, line 299: Make this the first one?

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,307 @@
...
+  stackdriver:  
+    project:  # stackdriver will use the default GCE credentials
+    metric_prefix: # empty to prevent the custom.googleapis.com/ prefix
+
+service:
+  pipelines:
+    # Remember to remove unwanted pipelines from this list
+    metrics/mysql:
+      receivers: [prometheus_exec/mysql]
+      processors: [filter/mysql, resourcedetection, metricstransform/mysql]
```
james-bebbington, line 286: Maybe make resourcedetection the last processor for all pipelines for consistency?

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,307 @@
...
+      exporters: [stackdriver]
+
+    metrics/apache:
+      receivers: [prometheus_exec/apache]
+      processors: [resourcedetection, metricstransform/apache]
+      exporters: [stackdriver]
+
+    metrics/host:
+      receivers: [hostmetrics]
+      processors: [resourcedetection, metricstransform/host]
```
james-bebbington, line 301: Also needs agentmetrics processor and filter similar to the config @JingboWangGoogle added on the Linux config PR

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,307 @@
...
+            aggregation_type: sum
+
+exporters:
+  stackdriver:  
+    project:  # stackdriver will use the default GCE credentials
+    metric_prefix: # empty to prevent the custom.googleapis.com/ prefix
+
+service:
+  pipelines:
+    # Remember to remove unwanted pipelines from this list
```
james-bebbington, line 283:
```suggestion
    # remove any unwanted pipelines from this list if desired
```

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,307 @@
...
+      network:
+      swap:
+
+  # https://github.com/prometheus/mysqld_exporter
+  prometheus_exec/mysql:
+  # --no-collect.global_variables reduces the amount of generated metrics, can be removed to enable all (600+) MySQL metrics
+    exec: ./prometheus_exporters/mysqld_exporter --web.listen-address=":{{port}}" --no-collect.global_variables
+    port: 9104
+    env:
+      # DATA_SOURCE_NAME and its value are needed to connect to the MySQL server, alternatively you could use a .my.cnf configuration file
```
james-bebbington, line 25: I don't have any context but is it standard to use a .my.cnf file? If so should we leave this commented out by default?

config/config-mysql_apache_jvm.yaml
```diff
@@ -0,0 +1,307 @@
...
+        operations:
+          # delete singular type dimension, retaining only direction
+          - action: aggregate_labels
+            label_set: [ direction ]
+            aggregation_type: sum
+
+exporters:
+  stackdriver:  
+    project:  # stackdriver will use the default GCE credentials
+    metric_prefix: # empty to prevent the custom.googleapis.com/ prefix
```
james-bebbington, line 279: Don't we need to add `custom.googleapis.com`? If we don't wouldn't the metrics get rejected?

What's the intention here btw? Any explicitly listed transformed metrics get sent to the corresponding agent metrics, and any other metrics are exported as custom metrics?

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: The exporter indicates to use either: https://github.com/prometheus/mysqld_exporter

The only extra config you need for the exporter to work is the `DATA_SOURCE_NAME`, so using it through an env variable I think is more convenient! I just let the user know they can use a config file if they'd like.

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: The idea here is that this config "mimicks" the agent and actually exports metrics in the same format as it would for third-party applications. So all the metrics in this file need to be prefixed by `agent.googleapis.com`. The reason why I hadn't just put the prefix in the `stackdriver` exporter was to allow some metrics to go through without that prefix, but I know now that you need a certain prefix for cloud moniroting. This means that within a certain receiver's metrics (say Java), I would need to filter out the "agent" metrics and put them in a separate pipeline from the "normal" ones. However, after discussions with Quentin, I will just whitelist the metrics wanted and ignore the rest, which means I can simply put them through the same exporter (I'll make an additional exporter which won't be used in this config but can be used by a user for custom metrics), and it will reduce the cost of additional metrics the user may not want. To recap I'll filter everything out but the metrics that are used to populate the pre-made dashboards on Cloud Monitoring which trigger automatically from the third-party metrics (JVM, MySQL and Apache)

config/config-mysql_apache_jvm.yaml
james-bebbington: Yea I didn't give very good context there but the main reason I asked is because its better if everything can work by default (pending env var setup). If possible it'd be great if there's nothing in the config file that a user NEEDS to change to get it workign

config/config-mysql_apache_jvm.yaml
james-bebbington: I think the intent to use the agent prefix for curated metrics and fall back to using the custom prefix for any other metrics is valid. To me it seems a bit limiting for users if they can't easily export any metrics that aren't on the list of curated agent metrics. I'm not sure that what you had in the config before was correct though, as I think you would need to explicitly add the custom prefix for any of the metrics that you don't have a transform set up - is that what happens when you set `metric_prefix` to blank? (I thought that would mean you don't set any prefix for non transformed metrics so they will end up not getting sent at all?).

But in terms of if it is important to let users export non-curated metrics, I'm not sure - Quentin would be more of an expert than me, so if he recommends doing it this way for now, then this LGTM.

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -266,42 +305,45 @@ processors:
...
         operations:
           # delete singular type dimension, retaining only direction
           - action: aggregate_labels
             label_set: [ direction ]
             aggregation_type: sum
 
 exporters:
-  stackdriver:  
+  stackdriver/agent:  
+    project:  # stackdriver will use the default GCE credentials
```
james-bebbington, line 317: I don't think that line is needed?

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -266,42 +305,45 @@ processors:
...
+    metric_prefix: agent.googleapis.com/
+
+  stackdriver/custom:  
     project:  # stackdriver will use the default GCE credentials
```
james-bebbington, line 321: Maybe explicitly put the `metric_prefix` here so its really obvious to people what's happening

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -128,45 +146,56 @@ processors:
...
       
       - metric_name: java_lang_Threading_ThreadCount
         action: update
-        new_name: agent.googleapis.com/jvm/thread/num_live
+        new_name: jvm/thread/num_live
       
       - metric_name: java_lang_OperatingSystem_OpenFileDescriptorCount
         action: update
-        new_name: agent.googleapis.com/jvm/os/open_files
+        new_name: jvm/os/open_files
```
james-bebbington, line 181: The above three metrics seem to be not be included in the filter but have transforms listed here, mistake?

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -1,6 +1,7 @@
 # This config mimicks the Cloud Monitoring Agent by collecting metrics from popular third-party services the Agent already supports
 # and formatting the metrics in the same way, so they appear similarly in Cloud Monitoring. The services targeted here are MySQL, Apache and JVM.
 # There is also support for DogStatsD metrics, with no transformations
+# Please configure your environment-specific details in the receivers part, especially for MySQL and JVM
 # Please remove/comment-out the pipelines you don't need/want in the last section.
 # Documentation specific to each exporter is found in the comment directly above its configuration
```
james-bebbington, line 7: Can you add a tab before line 22

config/config-mysql_apache_jvm_statsd.yaml
james-bebbington: Also would be good to make the order of the metrics listed in the filter match the order in the transforms just so it's a bit easier to read

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -214,16 +243,26 @@ processors:
       
       - metric_name: apache_sent_kilobytes_total
         action: update
-        new_name: agent.googleapis.com/apache/traffic
+        new_name: apache/traffic
         operations:
           - action: toggle_scalar_data_type
 
+  filter/host:
```
james-bebbington, line 250: I'd move the host metrics filter & transform processors to the top to match the order you have in the pipelines at the bottom of this file

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -214,16 +243,26 @@ processors:
...
+      exclude:
+        match_type: strict
+        metric_names:
+          - system.network.dropped_packets
+          - system.swap.usage
+          - system.filesystem.inodes.usage
+          - system.swap.page_faults
+
   metricstransform/host:
     transforms:
```
james-bebbington, line 261: Is there a reason we only have a few metrics here and not the whole list? (can copy from the config-linux file)

config/config-mysql_apache_jvm_statsd.yaml
james-bebbington: Nit: Also for the comments at the top of this file, it'd be better to wrap them at a regular line length rather than having some really long lines

README.md
```diff
@@ -7,6 +7,7 @@ This repository is focused on building and packaging the OpenTelemetry Collector
 
 To generate a tarball file that packages the executable OpenTelemetry binary with the configuration file compatible with Google Cloud Monitoring.
 1. Run `make build-tarball`
+    - To also include JVM, Apache, MySQL and StasD support, run `make build-tarball-exporters CONFIG_FILE=config-mysql_apache_jvm.yaml` instead
```
james-bebbington, line 10: It'd be good to make the `build-tarball-exporters` use that config file by default (but still allow it to be overridable) so this doesn't need to be part of the command users need to run.

config/config-mysql_apache_jvm_statsd.yaml
```diff
@@ -0,0 +1,349 @@
+# This config mimicks the Cloud Monitoring Agent by collecting metrics from popular third-party services the Agent already supports
```
james-bebbington, line 1: Maybe just call this `config-linux-with-exporters.yaml`?

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: You're right! Technically if the user has all their 3rd-party applications on default values and already has a `.my.cnf` file or the `DATA_SOURCE_NAME` environment variable, it could be run without opening the config file. I think it's highly unlikely, given there is also some JMX setup needed and one or more pipelines might be removed. I'll comment them out for now, since the user will probably read the file anyways, but the off-chance they run it knowing their `env` vars are already setup, my dummy `env` var would mess it up. So yeah, I'll comment it out!

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: Yeah I agree it's limiting, so what I'll do is a setup similar to the one I had before. I'll manually set those "curated" metrics to have the `agent.googleapis.com`, and it'll be up to the user to add their own prefix when renaming if they want more than the "curated" metrics. At least this way they're not stuck with the `agent.googleapis.com` prefix on all the metrics which would make cloud monitoring block the extra ones, if they're not those curated ones. Doesn't sound ideal either way, but at least this way the user CAN get those metrics, they just need to transform their name to include a prefix. It comes back to the same for the agent metrics but limits the users less for the additional metrics. Do you agree?

config/config-mysql_apache_jvm_statsd.yaml
Nicolas-MacBeth: Not 100% sure, but I'll remove. I can't really test anymore since I don't have any GCP credentials, but from the documentation seems like this should be OK. https://pkg.go.dev/contrib.go.opencensus.io/exporter/stackdriver/?tab=doc#Options

config/config-mysql_apache_jvm_statsd.yaml
Nicolas-MacBeth: nice catch! the rest should already be in order though!

config/config-mysql_apache_jvm_statsd.yaml
Nicolas-MacBeth: oh yeah, must not have taken the latest version by mistake, fixed!

config/config-mysql_apache_jvm.yaml
james-bebbington: I think you should:

- For all agent metrics, set the prefix specifically as part of the transform
- In the stackdriver exporter, specify the custom prefix, so that any non-transformed metrics will use that prefix by default
- Ensure that in the default config you are filtering out all metrics that are not explicitly transformed to agent metrics

That way all the user needs to do if they want to include additional custom metrics is to change the filter (they can transform the metrics too if desired, but they shouldn't have to do that).

config/config-mysql_apache_jvm_statsd.yaml
james-bebbington: I've never added it and it worked fine for me

config/config-mysql_apache_jvm.yaml
james-bebbington: I'm not sure about the second point I commented above actually, it's very hard to tell if that works as I expected from looking at the OpenCensus Go Exporter code. It would be nice if it did. Can you check? If not, then what you have is probably the best we can do for now.

Nicolas-MacBeth: Okay! sounds good, could you check Gitter, I sent you a message there! Thanks!

james-bebbington: Replied.

As for my comment above, actually it should work as I stated as per: https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/pull/273

I'm not sure if that was every released though. I'm going to release this now and then create a PR to update contrib. After that, maybe we can try updating.

Nicolas-MacBeth: Sent you another Gitter message (can't use Gchat sorry)

james-bebbington: Contrib was released with this change I believe. You can try updating the version of Contrib and testing if this now works (may help resolve your other dependency issue as well)

Nicolas-MacBeth:
@googlebot I fixed it.

config/config-mysql_apache_jvm.yaml
Nicolas-MacBeth: updated the config, but still getting the same build error. If you pull this branch, do you get the same thing?

README.md
```diff
@@ -74,6 +74,7 @@ Common configuration that you may want to change includes:
 To generate a tarball archive that includes the OpenTelemetry binary and a configuration file compatible with Google Cloud Monitoring:
 
 1. Run `make build-tarball`.
+  - To also include JVM, Apache, MySQL and StasD support, run `make build-tarball-exporters` instead
```
james-bebbington, line 77: Note four spaces represents an indent whereas two does not, so this will render better as (also fixes typo):
```suggestion
    - To also include JVM, Apache, MySQL and StatsD support, run `make build-tarball-exporters` instead
```

james-bebbington (changes requested):
CI failed with:
```
impi errors:
cmd/otelopscol/components.go: Import groups are not in the proper order: ["Third party" "Third party" "Local"] impi verification failed: Found 1 errors
Makefile:63: recipe for target 'impi' failed
```

Can you run `make presubmit` and fix any errors?

james-bebbington (approved):

PR #9, opened 2020-07-29
Added documentations as README

JingboWangGoogle:
1. A root directory README currently specifying the tarball generation process
2. A README in docs that will be packaged in the tarball file for the end user

### DIFF
--- a/.build/tar/generate_tar.sh
+++ b/.build/tar/generate_tar.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 CONFIG_FILE=${CONFIG_FILE:-config-example.yaml}
+FINAL_CONFIG_FILE=config.yaml
 README=${README:-tarball-readme.md}
 RUN=run.sh
 
@@ -27,10 +28,12 @@ cp bin/$OTELCOL_BINARY dist/
 cp docs/$README dist/
 cp .build/tar/$RUN dist/
 
+mv dist/$CONFIG_FILE dist/$FINAL_CONFIG_FILE
+
 # compress the binary and the config into a .tar file
 echo "Compressing..."
-cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $CONFIG_FILE $README $RUN
+cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $FINAL_CONFIG_FILE $README $RUN
 
 # remove the folders and files that were added
 echo "Clean up..."
-rm $OTELCOL_BINARY $CONFIG_FILE $README $RUN
+rm $OTELCOL_BINARY $FINAL_CONFIG_FILE $README $RUN
--- a/.build/tar/generate_tar.sh
+++ b/.build/tar/generate_tar.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 CONFIG_FILE=${CONFIG_FILE:-config-example.yaml}
+README=${README:-tarball-readme.md}
 
 echo "Start building tarball distribution file"
 
@@ -22,11 +23,12 @@ fi
 echo "Organizing files to be compressed"
 cp config/$CONFIG_FILE dist/
 cp bin/$OTELCOL_BINARY dist/
+cp docs/$README dist/
 
 # compress the binary and the config into a .tar file
 echo "Compressing..."
-cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $CONFIG_FILE
+cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $CONFIG_FILE $README
 
 # remove the folders and files that were added
 echo "Cleaning up"
-rm $OTELCOL_BINARY $CONFIG_FILE
+rm $OTELCOL_BINARY $CONFIG_FILE $README
--- a/.build/tar/generate_tar.sh
+++ b/.build/tar/generate_tar.sh
@@ -3,7 +3,6 @@
 CONFIG_FILE=${CONFIG_FILE:-config-example.yaml}
 FINAL_CONFIG_FILE=config.yaml
 README=${README:-tarball-readme.md}
-RUN=run.sh
 
 echo "Start building tarball distribution file..."
 
@@ -26,14 +25,13 @@ echo "Organizing files to be compressed..."
 cp config/$CONFIG_FILE dist/
 cp bin/$OTELCOL_BINARY dist/
 cp docs/$README dist/
-cp .build/tar/$RUN dist/
 
 mv dist/$CONFIG_FILE dist/$FINAL_CONFIG_FILE
 
 # compress the binary and the config into a .tar file
 echo "Compressing..."
-cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $FINAL_CONFIG_FILE $README $RUN
+cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $FINAL_CONFIG_FILE $README
 
 # remove the folders and files that were added
 echo "Clean up..."
-rm $OTELCOL_BINARY $FINAL_CONFIG_FILE $README $RUN
+rm $OTELCOL_BINARY $FINAL_CONFIG_FILE $README
--- a/.build/tar/generate_tar.sh
+++ b/.build/tar/generate_tar.sh
@@ -2,8 +2,9 @@
 
 CONFIG_FILE=${CONFIG_FILE:-config-example.yaml}
 README=${README:-tarball-readme.md}
+RUN=run.sh
 
-echo "Start building tarball distribution file"
+echo "Start building tarball distribution file..."
 
 # check config file
 if [ ! -e "config/$CONFIG_FILE" ]
@@ -20,15 +21,16 @@ then
 fi
 
 # move the needed files into dist folder
-echo "Organizing files to be compressed"
+echo "Organizing files to be compressed..."
 cp config/$CONFIG_FILE dist/
 cp bin/$OTELCOL_BINARY dist/
 cp docs/$README dist/
+cp .build/tar/$RUN dist/
 
 # compress the binary and the config into a .tar file
 echo "Compressing..."
-cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $CONFIG_FILE $README
+cd dist && tar -cvzf google-cloudops-opentelemetry-collector.tar.gz $OTELCOL_BINARY $CONFIG_FILE $README $RUN
 
 # remove the folders and files that were added
-echo "Cleaning up"
-rm $OTELCOL_BINARY $CONFIG_FILE $README
+echo "Clean up..."
+rm $OTELCOL_BINARY $CONFIG_FILE $README $RUN
--- a/.build/tar/run.sh
+++ b/.build/tar/run.sh
@@ -1,3 +1,0 @@
-echo "Thank you for using OpenTelemetry Collector and Google Cloud! You are currently running the binary with the default configuration file for Google Cloud Monitoring. If you want to add, edit, or remove receivers, processors, or exporters, please directly edit the file 'config-example.yaml'"
-
-./google-cloudops-opentelemetry-collector_linux_amd64 --config config-example.yaml
--- a/.build/tar/run.sh
+++ b/.build/tar/run.sh
@@ -1,0 +1,3 @@
+echo "Thank you for using OpenTelemetry Collector and Google Cloud! You are currently running the binary with the default configuration file for Google Cloud Monitoring. If you want to add, edit, or remove receivers, processors, or exporters, please directly edit the file 'config-example.yaml'"
+
+./google-cloudops-opentelemetry-collector_linux_amd64 --config config-example.yaml
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ checklicense:
 
 .PHONY: build
 build:
-	go build -o ./bin/$(OTELCOL_BINARY) ./cmd/otelopscols
+	go build -o ./bin/$(OTELCOL_BINARY) ./cmd/otelopscol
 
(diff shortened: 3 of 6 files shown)


docs/tarball-readme.md
```diff
@@ -0,0 +1,11 @@
+# OpenTelemetry Collector for Google Cloud Monitoring Tarball
+
+Thank you for using this OpenTelemetry Collector and Google Cloud Monitoring. This packaged tarball contains tools to immediately setup the Collector with compatible metrics for your Google Cloud Monitoring.
```
james-bebbington, line 3: Since we are not really going to be "officially" supporting this, can you say "Thank you for using this community build of the OpenTelemetry Collector for Google Cloud Monitoring..."

docs/tarball-readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Getting Started
+1. Run `bash run.sh` to run the collector
+2. Search for metrics in the Google Cloud Monitoring project, there should already be values populated!
+
+## Customed Metrics
```
james-bebbington, line 9:
```suggestion
## Custom Metrics
```

docs/tarball-readme.md
```diff
@@ -0,0 +1,11 @@
+# OpenTelemetry Collector for Google Cloud Monitoring Tarball
+
+Thank you for using this OpenTelemetry Collector and Google Cloud Monitoring. This packaged tarball contains tools to immediately setup the Collector with compatible metrics for your Google Cloud Monitoring.
+
+## Getting Started
+1. Run `bash run.sh` to run the collector
+2. Search for metrics in the Google Cloud Monitoring project, there should already be values populated!
```
james-bebbington, line 7: Is this true of the example config you include by default though? Should we create a full example config for Linux that actually generates all agent metrics?

docs/tarball-readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Customed Metrics
+
+Modify the configuration file `config-example` to experiement with different metrics
```
james-bebbington, line 11: How about as part of the tarball generation we rename this to just `config.yaml`?

README.md
```diff
@@ -0,0 +1,18 @@
+# OpenTelemetry Operations Collector
+
+This repository is focused on building and packaging the OpenTelemetry Collector with the necessary tools in different OSs to be compatible with Google Cloud Monitoring.
```
james-bebbington, line 3:
```suggestion
This repository is focused on building and packaging the OpenTelemetry Collector for use with Google Cloud Monitoring.
```

docs/tarball-readme.md
james-bebbington: > "This packaged tarball contains tools to immediately setup the Collector with compatible metrics for your Google Cloud Monitoring."

I'm not really sure what that part means. You also need to describe what these metrics are / where they come from.

README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Prerequisite
+
+[docker](https://docs.docker.com/engine/install/) is installed
+
+## Building the Docker Image
+
+Run `make docker-build-image`
+
+## Collector for Linux
```
james-bebbington, line 13: Using the docker image to run these make commands is useful if you want to replicate what the CI pipeline is doing directly, but most of the time users would just want to run the make commands directly imo. I'd probably have the initial example here call `make build-tarball`

Then add a brief paragraph at the end saying "alternatively you can run the make commands inside docker. First, generate a docker image for running the build commands by running `make docker-build-image` and then run `make TARGET=build-tarball docker-run`"

Makefile
```diff
@@ -86,4 +88,5 @@ docker-run:
 ifndef TARGET
 	$(error "TARGET is undefined")
 endif
+	@echo "Running $(TARGET) with docker container..."
```
james-bebbington, line 91:
```suggestion
	@echo "Running $(TARGET) in docker container $(BUILD_IMAGE_NAME)"
```

Makefile
```diff
@@ -38,7 +38,8 @@ OTELCOL_BINARY=google-cloudops-opentelemetry-collector
 
 .PHONY: build
 build:
-	go build -o ./bin/$(OTELCOL_BINARY)_$(GOOS)_$(GOARCH)$(EXTENSION) ./cmd/otelopscol
+	@echo "Building collector binary..."
```
james-bebbington, line 41: Do we need these? Doesn't make print out the command it's running anyway?

Makefile
```diff
@@ -28,7 +28,7 @@ ifeq ($(BUILD_IMAGE_NAME),)
 BUILD_IMAGE_NAME=otelopscol-build
 endif
 
-OTELCOL_BINARY=google-cloudops-opentelemetry-collector
+OTELCOL_BINARY=google-cloudops-opentelemetry-collector_$(GOOS)_$(GOARCH)$(EXTENSION)
```
james-bebbington, line 31: Same question as here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/6#discussion_r463359067

.build/tar/run.sh
```diff
@@ -0,0 +1,3 @@
+echo "Thank you for using OpenTelemetry Collector and Google Cloud! You are currently running the binary with the default configuration file for Google Cloud Monitoring. If you want to add, edit, or remove receivers, processors, or exporters, please directly edit the file 'config-example.yaml'"
```
james-bebbington, line 1: Not sure if we need this comment again? It's already in the README

.build/tar/generate_tar.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/bash
 
 CONFIG_FILE=${CONFIG_FILE:-config-example.yaml}
+README=${README:-tarball-readme.md}
```
james-bebbington, line 4: 👍

docs/tarball-readme.md
JingboWangGoogle: Let me know how it looks now.

docs/tarball-readme.md
JingboWangGoogle: Yes. I am planning on having a config file with all the host metrics, and all the transformations needed and all of them prefixed with agent.googleapis. I am currently working on that config file.

Makefile
JingboWangGoogle: Yes it does, but I find them not so readable, let me know if you find these unnecessary.

Makefile
JingboWangGoogle: I just tried it. First of all, it runs the build command with `windows`, and the final output in the dist folder is `google-cloudops-opentelemetry-collector.x86_64.0.0.0.goo`

.build/tar/run.sh
JingboWangGoogle: My goal here is to have a full coverage of comments like this. However, if you find this unnecessary, I can just remove the run.sh completely because there would only be one command here.

Makefile
james-bebbington: I dunno, it feels a bit overkill imo. The target name already describes what it does

Makefile
james-bebbington: Ok cool all good then

docs/tarball-readme.md
james-bebbington: It's the same isn't it? Did you forget to commit something

james-bebbington: It's the same isn't it. Did you forget to commit something?

.build/tar/run.sh
james-bebbington: Yea I dunno about this. I guess I don't mind too much either way, I just have a general preference for less stuff. In particular, less stuff that could possibly get out of sync (e.g. if someone built a tarball with a different config)

Also btw I think this is currently broken since you renamed the config file 😄

james-bebbington:
One other thing not particularly related to this PR that we should do is add a target to run `go test` and run this as part of all the `build-<package>` targets

docs/tarball-readme.md
JingboWangGoogle: Sorry, don't know what happened there. It is updated now :)

Makefile
JingboWangGoogle: That's true. Since it is mainly engineers who will be using this, I can see how these outputs don't necessarily help with this group of users. I just removed these.

.build/tar/run.sh
JingboWangGoogle: Ok, since there is only one command in here, I am not very strong in keeping this as well. For simplicity, I have just removed this file. If we ever see the need for adding this back, we can always do that.

README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This repository is focused on building and packaging the OpenTelemetry Collector for use with Google Cloud Monitoring.
+
+## Prerequisite
+
+[docker](https://docs.docker.com/engine/install/) is installed
+
+## Building the Docker Image
+
+Run `make docker-build-image`
```
james-bebbington, line 11:
```suggestion
```

README.md
```diff
@@ -0,0 +1,23 @@
...
+
+To generate a tarball file that packages the executable OpenTelemetry binary with the configuration file compatible with Google Cloud Monitoring.
+1. Run `make build-tarball`
+2. The tarball file will be generated in the `dist` folder
+
+Alternatively, you can  run the make commands inside docker.
+1. Make sure docker image has already been built: run `docker images` to check if `otelopscol-build` image exists
+2. If image doesn't exist, run `make docker-build-image` to build an image from the dockerfile
+2. Run `make TARGET=build-tarball docker-run`
+3. The tarball file will be generated in the `dist` folder
```
james-bebbington, line 23:
```suggestion
## Running build commands in Docker

Alternatively, you can run the build commands inside docker:
1. Run `make docker-build-image` to build the docker image. This will generate an image called `otelopscol-build`
2. Run `make TARGET=build-<package> docker-run`
3. The specified package will be generated in the `dist` folder
```

docs/tarball-readme.md
```diff
@@ -0,0 +1,11 @@
+# OpenTelemetry Collector for Google Cloud Monitoring Tarball
+
+Thank you for using this community build of the OpenTelemetry Collector for Google Cloud Monitoring. This packaged tarball contains tools to setup the Collector with some default host metrics for your Google Cloud Monitoring: memory, cpu, disk, filesystem, network, swap, process, load metrics. More details of these host metrics can be viewed [here](https://cloud.google.com/monitoring/api/metrics_agent). There metrics are scraped by the [host metric receiver](https://github.com/open-telemetry/opentelemetry-collector/tree/master/receiver/hostmetricsreceiver) in the collector.
```
james-bebbington, line 3:
```suggestion
Thank you for using this community build of the OpenTelemetry Collector for Google Cloud Monitoring. This tarball package contains the Collector configured to scrape system metrics from the host agent and report these to [Cloud Monitoring](https://cloud.google.com/monitoring/api/metrics_agent).
```

(trying to keep this very generic)

README.md
JingboWangGoogle: Do you think we don't need these here?

james-bebbington: I moved this section to the bottom in that other comment. You could leave the prerequisite section if you like.

config/config-example.yaml
```diff
@@ -4,27 +4,236 @@ receivers:
     collection_interval: 1m
     scrapers:
       memory:
+      cpu:
+      disk:
+      filesystem:
+      network:
+      swap:
+      process:
```
james-bebbington, line 12: None of the `process` metrics will work unless you use the custom processor that's pending the other PR: https://github.com/GoogleCloudPlatform/opentelemetry-operations-
collector/pull/2 (similar for a few of the other metrics)

I can chase up @nilebox to review that if this is a blocker for you

config/config-example.yaml
```diff
@@ -4,27 +4,236 @@ receivers:
...
+      load:
```
james-bebbington, line 13: For Linux there is also the `processes` scraper (this might need to be cleaned up a bit before its usable, but take a look if you have time)

config/config-example.yaml
```diff
@@ -4,27 +4,236 @@ receivers:
     collection_interval: 1m
     scrapers:
       memory:
```
james-bebbington, line 6: I would prefer if these were in alphabetical order 😄  (other than `process` that I usually put at the end since it is kind of a special case and a lot more expensive than the other scrapers)

config/config-example.yaml
james-bebbington: This is merged now if you want to update your config

JingboWangGoogle: I did update the config with some re-orderings. Is that what you are referring to? The `process` metric is already in the config.

james-bebbington: Some of the metrics you currently have configured to be scraped won't work unless I'm missing something, e.g.

- You have the `process` scraper configured. The OT process metrics store process related attributes in the resource object whereas Cloud Monitoring stores these as labels (https://cloud.google.com/monitoring/api/metrics_agent#agent-processes). This requires a fairly complex transformation not supported by the metrics transform processor so you need to add the `agentmetrics` processor to do that
- Likewise that processor takes care of splitting some metrics (disk_read, disk_write) & computing utilizations

config/config-example.yaml
```diff
@@ -1,39 +1,283 @@
...
+          - action: toggle_scalar_data_type
 
 exporters:
-  # export to Cloud Monitoring using the agent prefix
+  # export to cloud monitoring using the agent prefix
   stackdriver:
-    project: "otel-gce-metrics" # change this to the project name of the Monitoring project
+    metric_prefix: agent.googleapis.com/
+
+  # uncomment for debugging
```
james-bebbington, line 267:
```suggestion
  # to debug the outputted metrics, uncomment the following and add "logging" to the exporters list below
```

config/config-example.yaml
```diff
@@ -1,39 +1,283 @@
...
+      # system.swap.paging_ops -> swap/io
+      - metric_name: system.swap.paging_ops
+        action: update
+        new_name: swap/io
+        operations:
+          # delete singular type dimension, retaining only direction
+          - action: aggregate_labels
+            label_set: [ direction ]
+            aggregation_type: sum
+      # system.swap.usage -> pagefile/percent_used
```
james-bebbington, line 219: Comment is wrong (both left & right 😄 )

PR #11, opened 2020-08-12
Added "Collector for Windows" section to README

james-bebbington:


README.md
```diff
@@ -2,16 +2,24 @@
 
 This repository is focused on building and packaging the OpenTelemetry Collector for use with Google Cloud Monitoring.
 
-
 ## Collector for Linux
 
-To generate a tarball file that packages the executable OpenTelemetry binary with the configuration file compatible with Google Cloud Monitoring.
+To generate a tarball archive that includes the OpenTelemetry binary and a configuration file compatible with Google Cloud Monitoring:
```
JingboWangGoogle, line 7: This part might need to change after https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/6 merges, since this tarball will not just include a tarball and a configuration, but also a readme and some third party supports. Or maybe include this change in that PR?

james-bebbington: I think its okay to not mention the README file. As for the third party exporters, that is a different `make` command than the one listed here and README docs for that should probably be added as part of #6

JingboWangGoogle: included this change in #6

PR #27, opened 2020-11-09
Add IIS & MSSQL config

james-bebbington:


config/config-windows.yaml
```diff
@@ -26,7 +26,41 @@ receivers:
...
-  # agent self-observability metrics
+  # scrape iis perf counters at 1m interval
+  windowsperfcounters/iis:
+    collection_interval: 1m
+    perfcounters:
+      - object: Web Service
+        instances: _Total
+        counters:
+          # TODO: instead of setting capture group values to lowercase, make this option explicit
+          - Current Connections
```
nilebox, line 37: I suppose these names are defined somewhere as constants?
Might be useful to add a link in a comment.

config/config-windows.yaml
```diff
@@ -26,7 +26,41 @@ receivers:
...
+          - Total Connection Attempts (all instances)
+          - Total delete Requests
+          - Total get Requests
+          - Total head Requests
+          - Total options Requests
+          - Total post Requests
+          - Total put Requests
+          - Total trace Requests
+
+  # scrape mssql perf counters at 1m interval
```
nilebox, line 49: How does that work when MSSQL is not used / available in a particular use case? Will these counters be dropped?

nilebox (approved):

config/config-windows.yaml
james-bebbington: There counter names are defined by Microsoft. It's actually really hard to find a good link, but some relevant docs will be added here: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/receiver/windowsperfcountersreceiver#iis

config/config-windows.yaml
james-bebbington: Yea very good question. Currently it would crash.

I was actually just writing a PR to change this so that warnings are logged on startup instead which is probably a better experience.

james-bebbington:
Btw I don't plan to merge this anytime soon. Was mostly just storing this here so its easy for me to remember 😄

config/config-windows.yaml
james-bebbington: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/1538

igorpeshansky: +1.

igorpeshansky (approved):
LGTM :shipit:

PR #38, opened 2021-01-11
Fix string equality check in Makefile

ekund:
Check string equality with -eq to avoid 'unexpected operator' error from `make`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ EXTENSION := .exe
 endif
 
 # if ARCH is not supplied, set default value based on user's system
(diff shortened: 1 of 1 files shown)


ekund:
I am not following the shell style-guide which recommends to use '[[' instead of '[' for evaluations as per https://google.github.io/styleguide/shellguide.html#s6.3-tests

This is due to this script running on Windows and not assuming that there is bash installed, so it needs to work with /bin/sh for Linux.

igorpeshansky (approved):
LGTM :shipit:

PR #43, opened 2021-04-22
Update to Open Telemetry Collector 0.27

quentinmit:
NOTE: The `stackdriver` exporter has been renamed to `googlecloud`. Anyone using this will need to update their config files (the sample config files in this repo have been updated).


qingling128 (approved):
LGTM with a question:

It looks like the [upstream repo](https://github.com/open-telemetry/opentelemetry-collector-contrib) has deprecated the old [stackdriver exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/stackdriverexporter) in favor of the new [googlecloudexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlecloudexporter), so we are just migrating to the new exporter. Is this considered a backward incompatible change, in which case we need to bump the major version of this repo?

quentinmit:
> LGTM with a question:
> 
> It looks like the [upstream repo](https://github.com/open-telemetry/opentelemetry-collector-contrib) has deprecated the old [stackdriver exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/stackdriverexporter) in favor of the new [googlecloudexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlecloudexporter), so we are just migrating to the new exporter. Is this considered a backward incompatible change, in which case we need to bump the major version of this repo?

Well, I'll leave that decision to whoever next releases from this repo - I don't think I should be touching the version number as part of this PR.

qingling128:
There seems to be some bug with the license check bot. https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/cmd/otelopscol/main.go and https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/cmd/otelopscol/components.go have exactly the same license, but the bot complains about one but not the other. We can probably address the license thing in a followup PR, as it's broken on master as well.

qingling128:
> There seems to be some bug with the license check bot. https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/cmd/otelopscol/main.go and https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/cmd/otelopscol/components.go have exactly the same license, but the bot complains about one but not the other. We can probably address the license thing in a followup PR, as it's broken on master as well.

Ah, I see. The bot only checks files that have been touched by this PR. Either way, we could address the license complaints separately.

punya:
> NOTE: The `stackdriver` exporter has been renamed to `googlecloud`. Anyone using this will need to update their config files (the sample config files in this repo have been updated).

According to https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/stackdriverexporter, existing configs will continue to work for the sake of backwards compatibility. Does this repo do something special that breaks the compatibility measures from upstream?

qingling128:
Filed https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/issues/44 to track the license header check failures and records some initial findings.

punya:
Does this PR fully subsume #37? That PR also has some changes separate from the collector upgrade, such as tweaks to the Linux config files.

qingling128:
> > NOTE: The `stackdriver` exporter has been renamed to `googlecloud`. Anyone using this will need to update their config files (the sample config files in this repo have been updated).
> 
> According to https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/stackdriverexporter, existing configs will continue to work for the sake of backwards compatibility. Does this repo do something special that breaks the compatibility measures from upstream?

The options are:
1. Fully migrate to the new upstream `googlecloud` exporter
2. Support both `stackdriver` and `googlecloud` exporters for a period of time, and gradually deprecated the `stackdriver` one.

Since the main consumer of this repo is the Ops Agent that is fully under our control (plus it is an Alpha product + not officially supported by Google), 1 sounds more reasonable. I do think we need to bump the major version though.

image

> Does this PR fully subsume #37? That PR also has some changes separate from the collector upgrade, such as tweaks to the Linux config files.

There seems to be some additional metrics that changed. @quentinmit is updating this PR.

quentinmit:
> > NOTE: The `stackdriver` exporter has been renamed to `googlecloud`. Anyone using this will need to update their config files (the sample config files in this repo have been updated).
> 
> According to https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/stackdriverexporter, existing configs will continue to work for the sake of backwards compatibility. Does this repo do something special that breaks the compatibility measures from upstream?

As far as I can tell, the mentioned backwards compatibility is just plain broken.



> Does this PR fully subsume #37? That PR also has some changes separate from the collector upgrade, such as tweaks to the Linux config files.

I am not attempting to subsume #37, though I suspect it includes some of the same changes.

punya:
> Since the main consumer of this repo is the Ops Agent that is fully under our control (plus it is an Alpha product + not officially supported by Google), 1 sounds more reasonable.

Makes sense, thanks for explaining.

> As far as I can tell, the mentioned backwards compatibility is just plain broken.

It's troubling that the compatibility doesn't work -- @aabmass please take a look. (That doesn't need to impact this PR in any way, since we can just migrate to the new name. But we should probably fix it upstream to protect anybody else using the old name.)

quentinmit:
> > As far as I can tell, the mentioned backwards compatibility is just plain broken.
> 
> It's troubling that the compatibility doesn't work -- @aabmass please take a look. (That doesn't need to impact this PR in any way, since we can just migrate to the new name. But we should probably fix it upstream to protect anybody else using the old name.)

If someone is going to look at it, the problem is in the new "stackdriver" exporter's `go.mod` file - Go complains that it contains an invalid version number when I try to include it as a dependency of the Operations Collector.

aabmass:
> If someone is going to look at it, the problem is in the new "stackdriver" exporter's `go.mod` file - Go complains that it contains an invalid version number when I try to include it as a dependency of the Operations Collector.

I took a look. The OTel contrib repo is using a bunch of `replace` directives to reference local modules, which I did for stackdriverexporter to reference googlecloudexporter. I was able to get this PR to build with stackdriverexporter by doing

```
go mod edit -replace=github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudexporter@v0.0.0-00010101000000-000000000000=github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudexporter@v0.25.0
```

I'm not sure how we could actually fix that upstream? I'm also not a go expert, @punya wdyt? Is it worth pursuing?

punya:
@aabmass, the missing version problem you mentioned is something that seems widespread in collector-contrib. It occurs any time a module in that repo depends on another module defined in the same repo. Everything works out as long as you're building *in that repo* because of the replace directives, but anybody pulling in those modules as a dependency gets a rude surprise (as Quentin discovered).

The "right" fix would be for the release management tooling in collector-contrib to learn to fix version numbers whenever a release is tagged. I'm not sure that's a good use of our time right now, but it might be worth filing (or upvoting) an issue about it.

@quentinmit if you want to get the backwards compatibility measure to work in this repo, you'll need to add the same replace directive to the top-level go.mod. here It's up to you whether you want to do that, or just fix the YAML config files (and any internal source code that generates YAML files).

quentinmit:
> @aabmass, the missing version problem you mentioned is something that seems widespread in collector-contrib. It occurs any time a module in that repo depends on another module defined in the same repo. Everything works out as long as you're building _in that repo_ because of the replace directives, but anybody pulling in those modules as a dependency gets a rude surprise (as Quentin discovered).

I don't think it's a "missing version" so much as an "invalid version". I suspect if the dependency just specified `v0.0.0` instead of `v0.0.0-00010101000000-000000000000` it would work just fine.

punya:
@quentinmit is there anything blocking this PR? IIUC, the dependency version problem is a separate issue.

quentinmit:
> @quentinmit is there anything blocking this PR? IIUC, the dependency version problem is a separate issue.

I wanted to make sure that I wasn't introducing any new broken metrics before merging this PR. I think we might be there but I believe @davidbtucker is taking another pass over the full set of metrics to confirm.

igorpeshansky (approved):
LGTM :shipit:

PR #46, opened 2021-05-18
Added support for Nginx & added processor to normalize sum metrics for better ingestion into GCP

dehaansa:
Added the nginx receiver from collector-contrib.
Built a new processor to normalize sums/counters so that when ingested into GCP, they would use the first data point collected as a zero-point, to prevent large data spikes upon initial collection or restart of the agent.

This PR is also dependent on #43


dehaansa:
@googlebot I signed it!

processor/normalizesumsprocessor/processor.go
```diff
@@ -0,0 +1,369 @@
...
+		}
+		nsp.history[metricIdentifier] = &newStart
+
+		return false
+	}
+
+	// If this data is older than the start point, we can't
+	// meaningfully report this point
+	if dp.Timestamp() <= start.intDataPoint.Timestamp() {
+		return false
```
jsuereth, line 184: I'm not sure this is the right way to always handle this, e.g. in the event of a clock reset or other oddity.

Can you (at a minimum) add a TODO here to look into other mechanisms to handle time-rewinds and (at a maximum) find a way to remember we saw a time rewind, so if we see it a second time, we treat it like a roll-over/restart ?

If you want to see some other bugs of similar fashion, we had this one in OpenCensus: https://github.com/census-instrumentation/opencensus-java/issues/2068

processor/normalizesumsprocessor/processor.go
```diff
@@ -0,0 +1,369 @@
...
+			lastDoubleValue: newDP.Value(),
+		}
+		nsp.history[metricIdentifier] = &newStart
+
+		return false
+	}
+
+	// If this data is older than the start point, we can't
+	// meaningfully report this point
+	if dp.Timestamp() <= start.doubleDataPoint.Timestamp() {
```
jsuereth, line 241: same comment as above

processor/normalizesumsprocessor/processor.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	return dps.Len(), nil
+}
+
+func (nsp *NormalizeSumsProcessor) processIntSumDataPoint(dp pdata.IntDataPoint, resource pdata.Resource, metric *pdata.Metric) bool {
+	metricIdentifier := dataPointIdentifier(resource, metric, dp.LabelsMap())
+
+	start := nsp.history[metricIdentifier]
+	// If this is the first time we've observed this unique metric,
```
quentinmit, line 164: I believe this logic already exists in OT in the prometheus receiver; we should try to avoid duplicating that code here. (But also, it would be good to have some understanding of why nginxreceiver needs this in the first place - it sounds like the metrics it's emitting are already broken.)

processor/normalizesumsprocessor/config.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Metric defines the transformation applied to the specific metric
+type SumMetrics struct {
+	// MetricName is used to select the metric to operate on.
+	// REQUIRED
+	MetricName string `mapstructure:"metric_name"`
+
+	// NewName specifies the name of the new metric after transforming.
+	NewName string `mapstructure:"new_name"`
```
quentinmit, line 34: I think this is redundant with the metricstransformprocessor, why are we adding this functionality here as well?

quentinmit (changes requested):

processor/normalizesumsprocessor/config.go
dehaansa: That's a great point, the intention was to reduce the amount of processing required in the case that a metric did not require other processing via the transform processor, but it's certainly redundant.

processor/normalizesumsprocessor/processor.go
dehaansa: The metrics being emitted from the nginxreceiver have no start time reference due to the way metrics are exposed by nginx. In order to present meaningful data, we treat the first point as the start reference. You lose the information of the full counter value, but retain the information of how the counter value is growing over time.

processor/normalizesumsprocessor/processor.go
dehaansa: Went with the TODO and log. It wouldn't be very difficult to try to track the time-rewind, but adds some complexity and increases the size of the struct being tracked for each relevant metric. If you think that makes more sense, I'm happy to toss it in.

dehaansa:
@jsuereth @quentinmit addressed your feedback, but also pulled in some changes to the processor that we had implemented on other branches (mostly to resolve issues we discovered when metrics were provided in unexpected ways, like each labelled data point belonging to a different metric/slice)

djaglowski:
@jsuereth Can you advise on the `kokoro` check, what it is, why it doesn't seem to be running?

jsuereth:
@djaglowski Needed to add the "trust" label (currently requires a googler to do this)

processor/normalizesumsprocessor/processor.go
jsuereth: (a) the log line looks awesome. We might be able to create an alert if we see it often enough.
(b) My concern is on catastrophic failures (e.g. a time reset).   Would this receiver ever recover to a working state?  I think effectively it would fail to grab metrics and continue to do so.   That might be ok, but if there's something we can do to limp along reporting metrics, that's more ideal.

I think this is fine for now, because MOST failure scenarios I can imagine of this sort will either resolve or fail when actually sending to cloud monitoring due to the timestamp being wildly inaccurate.

processor/normalizesumsprocessor/processor.go
```diff
@@ -0,0 +1,336 @@
...
+	})
+	return b.String()
+}
+
+func addAttributeToIdentityBuilder(b *strings.Builder, v pdata.AttributeValue) {
+	switch v.Type() {
+	case pdata.AttributeValueTypeArray:
+		b.WriteString("[")
+		arr := v.ArrayVal()
+		for i := 0; i < arr.Len(); i++ {
```
jsuereth, line 282: FYI - We're likely to spec out this behavior so hopefully you get a helper method from the collector going forward.

djaglowski:
@jsuereth Will you please trigger kokoro again?

jsuereth:
```
golangci-lint run
level=warning msg="[runner] The linter 'scopelint' is deprecated (since v1.39.0) due to: The repository of the linter has been deprecated by the owner.  Replaced by exportloopref."
processor/normalizesumsprocessor/processor_test.go:190:2: commentFormatting: put a space between `//` and comment text (gocritic)
	//mb1.addIntDataPoint(1, map[string]string{}, startTime, 0)
	^
processor/normalizesumsprocessor/processor_test.go:198:2: commentFormatting: put a space between `//` and comment text (gocritic)
	//mb2.addIntDataPoint(5, map[string]string{}, startTime+2000, 0)
	^
processor/normalizesumsprocessor/processor_test.go:237:2: commentFormatting: put a space between `//` and comment text (gocritic)
	//mb1.addIntDataPoint(1, map[string]string{"label": "val2"}, startTime+2000, 1)
	^
processor/normalizesumsprocessor/processor_test.go:476:2: singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
	switch t {
	^
Makefile:80: recipe for target 'lint' failed
make: *** [lint] Error 1
```

dehaansa:
@jsuereth linter issues fixed, can we try again?

jsuereth (approved):

processor/normalizesumsprocessor/processor.go
quentinmit: Yeah, I'm just trying to argue against duplicated code - it seems like this is something that should (eventually) be fixed in the nginxreceiver directly, using the same logic and code as the prometheus receiver already does. There's lots of subtle edge cases to handle, as I'm sure you've discovered, so it's better to fix it once right and carefully.

(or is this already a fork of the code in the prometheus receiver?)

config/config-windows.yaml
```diff
@@ -172,6 +172,9 @@ processors:
         operations:
           # change data type from int64 -> double
           - action: toggle_scalar_data_type
+        operations:
+          # change data type from int64 -> double
+          - action: toggle_scalar_data_type
```
quentinmit, line 177: Merge error?

otelcol.yaml
```diff
@@ -0,0 +1,25 @@
+receivers:
```
quentinmit, line 1: You probably didn't mean to commit this test file (and if you did, please don't put it in the root of the repo)

quentinmit (commented):
Left a couple comments but I'd like to figure out the high-level story around the duplicated logic before I look too hard at the code.

processor/normalizesumsprocessor/processor.go
quentinmit: Let me be clearer on this:

I know from personal experience that this is very subtle logic that is hard to get right in all the edge cases.

AFAIK this logic already exists in the OT collector repo, as the type [`MetricsAdjuster`](https://github.com/open-telemetry/opentelemetry-collector/blob/6e5beaf43b325e63ec6f1e864d9746a0d051cc35/receiver/prometheusreceiver/internal/metrics_adjuster.go#L187). Long-term I'd like to see this logic exist in exactly one place in OT, with a robust set of tests, called by all the different receivers/processors/exporters that need it. I think it's more likely that the logic will be correct if we do that.

I'm okay with making cleaning this up a TODO for now, since I don't want to block on this, but I don't want to end up in this state long-term. Can you add a TODO comment to merge this with `MetricsAdjuster` in a single common location? Then I'll do my best to review the current logic for any subtle bugs, but as I said, I know how hard it is to get right.

processor/normalizesumsprocessor/processor.go
```diff
@@ -0,0 +1,336 @@
...
+func (nsp *NormalizeSumsProcessor) transformMetrics(rms pdata.ResourceMetrics) []error {
+	var errors []error
+
+	ilms := rms.InstrumentationLibraryMetrics()
+	for j := 0; j < ilms.Len(); j++ {
+		ilm := ilms.At(j).Metrics()
+		newSlice := pdata.NewMetricSlice()
+		for k := 0; k < ilm.Len(); k++ {
+			metric := ilm.At(k)
+			if nsp.shouldTransformMetric(metric) {
```
quentinmit, line 76: Now that there's no renaming happening, do we need to have any configuration at all, or can we juts process any metric that doesn't have start times defined?

quentinmit:
I don't see any logic errors at the moment, so I am okay to approve this with a TODO to clean up the duplication.

otelcol.yaml
dehaansa: Thanks, good catch.

processor/normalizesumsprocessor/processor.go
dehaansa: There's currently two modes, one for do all sum metrics, one for targeting specific sum metrics.

There is no (to my understanding) "undefined" start time, are we comfortable stating that anything with a 0 start time (so epoch time, start of 1970) is undefined and would be modified by this? Seems like a reasonable assumption, just wanted to confirm.

processor/normalizesumsprocessor/processor.go
dehaansa: That's reasonable. In a quick look, it appears that MetricsAdjuster in its current state does not attempt to use the initial collection as a zero value for a sum metric, but otherwise it's definitely a lot of duplication and shared concepts. Happy to add the TODO .

processor/normalizesumsprocessor/processor.go
jsuereth: Yeah, that's a reasonable assumption.  This is proto3, so it defaults to 0, but also we can assume that we shouldn't be looking at programs that started in 1970.

quentinmit (approved):

processor/normalizesumsprocessor/README.md
```diff
@@ -0,0 +1,32 @@
...
+Additionally, any time a metric value decreases, it will be assumed to
+have rolled over, at which point another data point will be skipped in
+favor of providing only accurate data representing the sum from a known point
+
+## Configuration
+
+List the metrics being transformed under 'transforms'.
+
+```yaml
+normalizesums:
```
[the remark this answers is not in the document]
dehaansa: Thanks, missed that when refactoring out the config.

PR #49, opened 2021-06-29
Add a system.disk.average_utilization_time metric

quentinmit:
This represents the fraction of time that a disk is busy performing I/O. This is the underlying value needed for `agent.googleapis.com/disk/operation_time`

### DIFF
--- a/processor/agentmetricsprocessor/utils_calculate_average_disk.go
+++ b/processor/agentmetricsprocessor/utils_calculate_average_disk.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/processor/agentmetricsprocessor/utils_calculate_average_disk.go
+++ b/processor/agentmetricsprocessor/utils_calculate_average_disk.go
@@ -71,10 +71,14 @@ func (mtp *agentMetricsProcessor) appendAverageDiskMetrics(rms pdata.ResourceMet
 					continue
 				}
 			}
+			if len(newOp) == 0 {
+				// No point making a new metric if there is no data.
+				continue
+			}
 			averageTimeMetric := pdata.NewMetric()
 			opTimeMetric.CopyTo(averageTimeMetric)
 			opTimeMetric.SetName(metricPostfixRegex.ReplaceAllString(opTimeMetric.Name(), "average_operation_time"))
-			ddps := opTimeMetric.DoubleGauge().DataPoints()
+			ddps := opTimeMetric.DoubleSum().DataPoints()
 			ddps.Resize(len(newOp))
 			i := 0
 			for key, new := range newOp {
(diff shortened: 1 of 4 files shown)


processor/agentmetricsprocessor/utils_calculate_average_disk.go
```diff
@@ -0,0 +1,110 @@
...
+				default:
+					continue
+				}
+			}
+			if len(newOp) == 0 {
+				// No point making a new metric if there is no data.
+				continue
+			}
+			averageTimeMetric := pdata.NewMetric()
+			opTimeMetric.CopyTo(averageTimeMetric)
```
dashpole, line 83: What do we actually end up keeping from the opTimeMetric, and what do we overwrite?  Is there a reason why we started with opTime, rather than just operations?

processor/agentmetricsprocessor/utils_calculate_average_disk.go
```diff
@@ -0,0 +1,110 @@
...
+			opTimeMetric.SetName(metricPostfixRegex.ReplaceAllString(opTimeMetric.Name(), "average_operation_time"))
+			ddps := opTimeMetric.DoubleSum().DataPoints()
+			ddps.Resize(len(newOp))
```
dashpole, line 86: I'm slightly confused here.  Do we expect them to generally be the same size (is this just a precaution?).  Is this always going to truncate ddps?  If it isn't always going to truncate, will the newly added empty points still end up with all of the correct information? If that is the case, why do we bother starting with opTimeMetric?

processor/agentmetricsprocessor/utils_calculate_average_disk.go
quentinmit: Starting with opTime means the metric has the right unit, kind, description, etc. Operations is an int metric with a different unit. We only overwrite the name and the values.

dashpole: Do we need to overwrite the description as well?

processor/agentmetricsprocessor/utils_calculate_average_disk.go
quentinmit: I added this Resize as a precaution; as currently implemented I do think they're the same size, but there could be cases in the future where they're not (e.g. what do we do if the start timestamp resets? Should we swallow the first point on a stream or should we assume it all happened in the most recent period? Etc.)

processor/agentmetricsprocessor/utils_calculate_average_disk.go
quentinmit: In theory we could but nothing consumes the description and none of our other processors do.

dashpole (approved):

processor/agentmetricsprocessor/utils_calculate_average_disk.go
quentinmit: (I copied this pattern from e.g. utils_calculate_utilization.go:184)

qingling128 (approved):
LGTM

qingling128 (approved):
LGTM

PR #51, opened 2021-07-01
Unfork OT's metricstransformprocessor

quentinmit:
This replaces the OT fork with a tiny addition to agentmetricsprocessor to allow adding a blank label to arbitrary metrics.


qingling128 (approved):
LGTM

processor/agentmetricsprocessor/agentmetricsprocessor.go
```diff
@@ -78,6 +80,11 @@ func (mtp *agentMetricsProcessor) ProcessMetrics(ctx context.Context, metrics pd
 		errors = append(errors, err)
 	}
 
+	// Add blank labels last so it can be used on newly-generated metrics.
```
igorpeshansky, line 83: Can you please expand on this comment? What is the connection between newly-generated metrics and adding the blank label last?

processor/agentmetricsprocessor/config.go
```diff
@@ -19,4 +19,8 @@ import "go.opentelemetry.io/collector/config"
 // Config defines configuration for Resource processor.
 type Config struct {
 	config.ProcessorSettings `mapstructure:",squash"`
+
+	// BlankMetrics is a list of metrics that need a label called
+	// "blank" with an empty value.
+	BlankMetrics []string `mapstructure:"blank_metrics"`
```
igorpeshansky, line 25: `BlankLabelMetrics`?

processor/agentmetricsprocessor/config.go
```diff
@@ -19,4 +19,8 @@ import "go.opentelemetry.io/collector/config"
 // Config defines configuration for Resource processor.
 type Config struct {
 	config.ProcessorSettings `mapstructure:",squash"`
+
+	// BlankMetrics is a list of metrics that need a label called
+	// "blank" with an empty value.
```
igorpeshansky, line 24: This seems like a hack. Would this even work if you ever wanted to add multiple labels with blank values?
Is this done solely to avoid a fork? Are we expecting to undo this as soon as upstream allows blank values in the `add_label` transform?

config/config-example.yaml
```diff
@@ -36,7 +36,7 @@ processors:
   resourcedetection:
     detectors: [gce]
 
-  # perform custom transformations that aren't supported by the googlemetricstransform processor
+  # perform custom transformations that aren't supported by the metricstransform processor
```
igorpeshansky, line 39: [Optional] Do we want to add an example of adding a blank metric label?

igorpeshansky (changes requested):

processor/agentmetricsprocessor/agentmetricsprocessor.go
quentinmit: I'm not sure what other words I can use to explain this; agentmetricsprocessor creates new metrics, and we need to be able to add a blank label to one of those metrics.

processor/agentmetricsprocessor/config.go
quentinmit: Of course this is a hack. It's a much smaller hack than the previous hack. This will go away as soon as `add_label` is fixed upstream, just like most of the other things in agentmetricsprocessor.

config/config-example.yaml
quentinmit: No, because this is a temporary hack that no-one should use.

jsuereth (approved):

processor/agentmetricsprocessor/agentmetricsprocessor.go
igorpeshansky: How about: `// Add blank labels last so they also apply to metrics added by agentmetricsprocessor.`?

igorpeshansky (approved):
LGTM :shipit:

processor/agentmetricsprocessor/agentmetricsprocessor.go
quentinmit: Done.

PR #53, opened 2021-08-19
Add github action for validation with public logs.

jsuereth:
This does most of what kokoro does, but we'll play with it over time to get it right.


djaglowski:
@jsuereth Looks like this error is not directly related to your changes, except that you've made it very easy to spot a preexisting issue.

The issue is with the way tooling dependencies are being managed. Basically, `make install-tools` affects your `go.mod` and `go.sum`. You can see this on `master`:

```
git status
// clean
go mod tidy
// clean
make install-tools
// go.mod and go.sum in bad state
go mod tidy
// go.mod and go.sum fixed
```

So the easiest way to fix this, is to just run `make install-tools && go mod tidy` and then commit the resulting `go.mod` and `go.sum`. The problem is that it is very easy for this to happen again, because someone who has not run `make install-tools` will end up in a similar state as you are here.

Typically, it's better to move your tooling dependencies into an internal module, so they are isolated from your code. I opened #54 to isolate the tooling dependencies. Feel free to merge it to this branch, or if you'd prefer, make the easy fix and I'll rebase it on master.

PR #54, opened 2021-08-24
Move tools into internal module

djaglowski:


Makefile
```diff
@@ -43,11 +45,11 @@ LD_FLAGS := -ldflags "${BUILD_X1} ${BUILD_X2}"
 
 .PHONY: install-tools
 install-tools:
-	go get github.com/client9/misspell/cmd/misspell
-	go get github.com/golangci/golangci-lint/cmd/golangci-lint
-	go get github.com/google/addlicense
-	go get github.com/google/googet/goopack
-	go get github.com/pavius/impi/cmd/impi
+	cd $(TOOLS_DIR) && go install github.com/client9/misspell/cmd/misspell
```
quentinmit, line 48: Does switching from `go get` to `go install` change the behavior when the binary is already installed? I think I would be confused if I ran `make install-tools` and the binaries in my home directory were silently downgraded to the versions pinned in `go.mod` (if I understand the behavior of the new calls correctly).

jsuereth: I'm not sure this is as important as keeping the module file clean.

Specifically, the technique of make install-tools you see here is pretty stale compared to the new technique used in OTel for Go modules, which matches this "install-tools" technique.

To some extent, I think install-tools SHOULD upgrade/downgrade install tools so CI is consistent.  To another extent, we may want to hide the installation of tools inside a docker build if we want repeatable (and consistent) style checks.

jsuereth (approved):

PR #56, opened 2021-10-01
Upgrade to Open Telemetry 0.36

quentinmit:
This required a massive refactoring because of the changed metrics model. Hopefully we have good test coverage, because there were a ton of bugs after the refactoring that I had to track down with the tests.


cmd/otelopscol/main.go
```diff
@@ -54,7 +55,7 @@ func runInteractive(params service.CollectorSettings) error {
 		return fmt.Errorf("failed to construct the application: %w", err)
 	}
 
-	err = app.Run()
+	err = app.Run(context.Background())
```
djaglowski, line 58: The upstream has inserted a wrapper method [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/otelcol/main.go#L46:L47), which may be worth including in this update.

processor/normalizesumsprocessor/processor.go
```diff
@@ -97,229 +93,96 @@ func (nsp *NormalizeSumsProcessor) transformMetrics(rms pdata.ResourceMetrics) [
...
-func dataPointIdentifier(resource pdata.Resource, metric pdata.Metric, labels pdata.StringMap) string {
+func dataPointIdentifier(resource pdata.Resource, metric pdata.Metric, labels pdata.AttributeMap) string {
 	var b strings.Builder
 
 	// Resource identifiers
 	resource.Attributes().Sort().Range(func(k string, v pdata.AttributeValue) bool {
-		fmt.Fprintf(&b, "%s=", k)
-		addAttributeToIdentityBuilder(&b, v)
-		b.WriteString("|")
+		fmt.Fprintf(&b, "%s=%s|", k, v.AsString())
```
dehaansa, line 177: The only potential issue with v.AsString() is (last I looked) it doesn't sort map type attribute values, which could lead to inconsistent data point identifiers. This is pretty edge case, but when I was looking at refactoring normalizesums for 0.36 I noticed this. Should we be concerned about this edge case?

dehaansa: Nevermind, it looks like it's using json.marshal which should produce consistent order. 👍

quentinmit:
Ridwan ran the metric comparison script and it looks like there's a few metrics that don't exists under this collector. I'm guessing either metrics have been renamed upstream or have been removed/disabled and we need to turn them back on. He's going to post a list in this PR.

cmd/otelopscol/main.go
quentinmit: That code you linked to is not in 0.36 yet, unfortunately.

djaglowski: [Here](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.36.0/cmd/otelcol/main.go) is the v0.36.0 tagged version of it. Slightly different usage, but the `service` package appears to be present.

ridwanmsharif:
Here is a list of metrics that don't seem to be reported by the otel agent that was being reported by the existing otel subagent for ops-agent:

- agent.googleapis.com/agent/api_request_count
- agent.googleapis.com/cpu/utilization
- agent.googleapis.com/cpu/load_15m
- agent.googleapis.com/cpu/load_1m
- agent.googleapis.com/cpu/load_5m
- agent.googleapis.com/disk/bytes_used
- agent.googleapis.com/disk/io_time
- agent.googleapis.com/disk/merged_operations
- agent.googleapis.com/disk/operation_count
- agent.googleapis.com/disk/operation_time
- agent.googleapis.com/disk/pending_operations
- agent.googleapis.com/disk/percent_used
- agent.googleapis.com/disk/read_bytes_count
- agent.googleapis.com/disk/weighted_io_time
- agent.googleapis.com/disk/write_bytes_count
- agent.googleapis.com/interface/errors
- gent.googleapis.com/interface/packets
- gent.googleapis.com/interface/traffic
- agent.googleapis.com/memory/bytes_used
- agent.googleapis.com/memory/percent_used
- agent.googleapis.com/network/tcp_connections
- agent.googleapis.com/swap/bytes_used
- agent.googleapis.com/swap/io

Here is the list of metrics that are reported by both agents:
- agent.googleapis.com/agent/memory_usage
- agent.googleapis.com/agent/monitoring/point_count (the ops-agent version has two statuses: INVALID_ARGUMENT and OK where as v0.36 just has OK?)
- agent.googleapis.com/agent/uptime

Here are a few metrics that neither reported but the test expected:
- agent.googleapis.com/cpu/usage_time
- agent.googleapis.com/swap/percent_used


Used a debian-10 buster image for the VM and used the ops-agent latest release for one otel subagent and  for the second one we built it from this branch. The parity script didn't run (WIP) because of some permission issues, but the list above is what I collected by manually checking each metric for the vm.

quentinmit:
Given that the only metrics you found are `agent.googleapis.com/agent/` and those are in a separate pipeline, I would suspect that there's something wrong with your config file that's causing the other pipeline not to run or not to write to GCM with the correct resource labels. Can you add a loggingexporter to the pipeline and find out where the metrics are getting dropped?

ridwanmsharif:
I got the parity test script running and here are some findings:

I'm seeing some disk metrics being reported: agent.googleapis.com/disk/percent_used, agent.googleapis.com/disk/bytes_used, agent.googleapis.com/disk/read_bytes_count for example are being reported for my latest run. When fixing up and running the parity script I found that they were again unreported and instead agent.googleapis.com/cpu/load_15m, agent.googleapis.com/cpu/load_5m, agent.googleapis.com/cpu/load_1m were being reported (and passed the test)

The logging exporter added didn't show much, it only logged datapoints for the metrics being reported and logged nothing that I was expecting but went unreported. These findings suggest a code issue in the branch more than a configuration issue.

I also found from multiple runs of the subagent that its not the same metrics that are unreported. In some runs the CPU metrics get reported and in other ones the disk ones are reported. Seems very non-deterministic

ridwanmsharif:
We found a bug and the latest commit adds the fix for it. Metrics were being filtered out incorrectly (the code was overwriting the slice of metrics instead of appending to it) which caused metrics to not be processed. The parity test was run after the latest change and 11 of the metrics pass but the other 17 still fail. This failure seems to be more because the margin of error on the parity test seems to not be appropriate for certain tests. Will tweak the test and comment here when there are updates.

quentinmit:
The parity test script was reporting a lot of `NaN`s and spot checks of some of the listed metrics show ~identical values. I don't think any of the 17 failures are necessarily real. (But I agree we should investigate the test and wait for a passing run before we merge this.)

ridwanmsharif:
Parity tests now work and allow some margin of error when comparing metric values. Roughly 25/28 (sometimes some of the cpu usage ones fail too) pass the test have report metrics within 5% of the previous version. Some metrics however don't fall within this margin. Below are an example from one run of the parity test. I don't know if these differences block the change from being merged. Unclear what the degree of parity we are looking for is.

```
FAILED ParityMetricsTest::test_disk_operation_time (agent.googleapis.com/disk/operation_time: old = 13158.00, new = 2807.00)
FAILED ParityMetricsTest::test_network (agent.googleapis.com/network/tcp_connections: old = 15.00, new = 32.00)
FAILED ParityMetricsTest::test_cpu_utilization (agent.googleapis.com/cpu/utilization: old = 0.00, new = 0.02)
```

quentinmit:
Is `new` the OT 0.36 version? `cpu/utilization` is fine, `0.02` is essentially `0.00`.

`network/tcp_connections` conceivably could be off due to aliasing, since they're not polling at the same time. Can you check the graph by hand to see if they're roughly the same or not?

`disk/operation_time` is somewhat concerning. What aligner is being applied to that metric?

ridwanmsharif:
Yes new is v0.36 and old is the latest release of the ops-agent.

For `disk/operation_time` the graph looks like the following:
image
They roughly match each other, not sure what degree of similarity we're looking for. But it does look like maybe comparing the last value might not make sense for metrics like these. Not sure what exactly you mean by aligner but both configs (for old and new) use the same config and the tests don't add any offset.

For `network/tcp_connections` the graphs look like the following:
image
They are roughly the same but there are minor differences (for example for old the tcp_state=ESTABLISHED label is consistently at 15 but for new its at 14) the other differences look like they can be explained away by not polling at the same time

quentinmit:
Both of those look totally fine.

By "aligner" I mean the value of `aggregation.perSeriesAligner` in the `ListTimeSeriesRequest` that it sends.

qingling128 (approved):
Thanks @quentinmit and @ridwanmsharif for getting the parity tests to work and addressing the discrepancies it revealed!

The PR LGTM with a few optional comments / questions.

processor/normalizesumsprocessor/processor.go
```diff
@@ -97,229 +93,96 @@ func (nsp *NormalizeSumsProcessor) transformMetrics(rms pdata.ResourceMetrics) [
...
-
-		return false
-	}
-
-	start.lastDoubleValue = dp.Value()
-	dp.SetValue(dp.Value() - start.doubleDataPoint.Value())
-	dp.SetStartTimestamp(start.doubleDataPoint.Timestamp())
-
-	return true
+	return false, nil
```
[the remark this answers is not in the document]
quentinmit: Yeah, I think you're right that it could be. Because of the refactoring I removed the redundant check of the metric's type that was here, so there's no longer any error conditions. I didn't change the function signature to keep the diff smaller, but by removing this I was able to clean up a whole tree of error handling, so done now.

processor/agentmetricsprocessor/utils_calculate_average_disk.go
```diff
@@ -36,43 +36,32 @@ func (mtp *agentMetricsProcessor) appendAverageDiskMetrics(rms pdata.ResourceMet
...
-						if !ok {
-							op = mtp.prevOp[key]
-						}
-						// Can't just save idp because it is overwritten by OT.
-						op.operations = pdata.NewIntDataPoint()
-						idp.CopyTo(op.operations)
-						newOp[key] = op
-					}
 				case opTimeName:
 					opTimeMetric = metric
```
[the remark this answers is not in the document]
qingling128: Oh that makes sense!

processor/agentmetricsprocessor/utils_split_read_write.go
```diff
@@ -92,64 +92,38 @@ func splitReadWriteBytesMetric(metric pdata.Metric) (read pdata.Metric, write pd
...
+func appendNumberDataPoints(metricName string, ndps, read, write pdata.NumberDataPointSlice) error {
+	for i := 0; i < ndps.Len(); i++ {
+		ndp := ndps.At(i)
+		labels := ndp.Attributes()
 
 		dir, ok := labels.Get(directionLabel)
 		if !ok {
 			return fmt.Errorf("metric %v did not contain %v label as expected", metricName, directionLabel)
 		}
-		labels.Delete(directionLabel)
```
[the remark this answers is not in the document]
quentinmit: It moved to (new) line 126.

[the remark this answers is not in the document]
qingling128: Oh NVM.

qingling128 (approved):
LGTM

PR #58, opened 2021-10-07
Ensure the sums normalizer only acts on Monotonic sums

dehaansa:
It was recently discussed the non-monotonic sums are intended to be used like gauges that can be summed together, which are not metrics that need to be normalized by this processor. ~~Also removed some leftover test logging from the v0.36 branch.~~


jsuereth (approved):

processor/agentmetricsprocessor/agentmetricsprocessor_test.go
```diff
@@ -108,11 +107,6 @@ func TestAgentMetricsProcessor(t *testing.T) {
 			err = rmp.ConsumeMetrics(context.Background(), tt.input)
 			require.NoError(t, err)
 
-			marshaler := otlp.NewJSONMetricsMarshaler()
```
quentinmit, line 111: This was not leftover - this was intentionally left in, for the benefit of future debuggers.

(Note that test log output is not shown unless the test fails or you pass -v)

quentinmit (changes requested):
LGTM except for the logging change.

PR #59, opened 2021-10-18
Update Collector to v0.37.0

Mrod1598:
Updating the collector is required to further MySQL and Apache Metrics receivers.


cmd/otelopscol/main.go
```diff
@@ -49,11 +49,11 @@ func main() {
 }
 
 func runInteractive(params service.CollectorSettings) error {
-	app, err := service.New(params)
+	_, err := service.New(params)
```
quentinmit, line 52: I think this is redundant now - why call `service.New` if we're not going to use its output?

BTW I don't know if landed in 0.37 but there's a change on HEAD that will be backwards incompatible: https://github.com/open-telemetry/opentelemetry-collector/pull/4140

quentinmit: Yes, if it happens to any of the metrics that we map to `agent.googleapis.com/agent/...`

Mrod1598: so it is on 37. https://github.com/open-telemetry/opentelemetry-collector/blob/028fc46e667e2673b93d39372825aee5036eaa98/service/telemetry.go#L48
what does that mean for this PR? what updated is needed to get this through?

quentinmit: Why the processor and not Ops Agent config?

PR #68, opened 2021-12-14
cmd/otelopscol: Filter intrusive log spam

braydonk:
As outlined in open-telemetry/opentelemetry-collector#3004 the Open Telemetry Collector writes a consistent error log that is not user actionable and we would not like to see. This PR adds a logging option configuration that will detect the spamming log message and filter it.

### DIFF
--- a/cmd/otelopscol/main.go
+++ b/cmd/otelopscol/main.go
@@ -46,29 +46,32 @@ func main() {
 		Version:     version.Version,
 	}
 
-	// Remove hostmetrics logspam
-	logFilterFunc := func(entry zapcore.Entry, fields []zapcore.Field) bool {
-		if strings.Contains(entry.Caller.File, "scrapercontroller.go") {
-			if strings.Contains(entry.Message, "error reading process name for pid") {
-				return false
-			}
-		}
-		return true
-	}
-
 	params := service.CollectorSettings{
 		Factories: factories,
 		BuildInfo: info,
 		LoggingOptions: []zap.Option{
-			zap.WrapCore(func(core zapcore.Core) zapcore.Core {
-				return zapfilter.NewFilteringCore(core, logFilterFunc)
-			}),
+			logSpamFilterCore(),
 		},
 	}
 
 	if err := run(params); err != nil {
 		log.Fatal(err)
 	}
+}
+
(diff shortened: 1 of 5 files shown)


cmd/otelopscol/main.go
```diff
@@ -41,13 +46,34 @@ func main() {
...
 		log.Fatal(err)
 	}
 }
 
+// Returns a zapfilter core that will filter log spam from the otel collector.
+// Upstream issue: https://github.com/open-telemetry/opentelemetry-collector/issues/3004
+func logSpamFilterCore() zap.Option {
+	logFilterFunc := func(entry zapcore.Entry, fields []zapcore.Field) bool {
+		if strings.Contains(entry.Caller.File, "scrapercontroller.go") {
+			return strings.Contains(entry.Message, "error reading process name for pid")
```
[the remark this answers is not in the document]
braydonk: You're right, and looking into this made me realize that this code doesn't do what I thought it does. Working on a fix.

cmd/otelopscol/main.go
```diff
@@ -41,13 +45,45 @@ func main() {
...
+		}
+		for _, field := range fields {
+			if field.Key == "error" {
+				logError, ok := field.Interface.(error)
+				if !ok {
+					return true
+				}
+				return !strings.Contains(
+					logError.Error(),
+					"error reading process name for pid")
```
qingling128, line 76: Can we make this a constant slice (aka a list of strings that store the error messages we want to exclude)? The benefit is that it's easier to add additional filter rules.

e.g.

```
func logMessagesToExclude() []string {
    return []string{
        "error reading process name for pid",
    }
}
```

quentinmit: If you do, make it a global variable, not a func.

braydonk: I made this a more generic abstraction that will accept a list of error message substrings and made the comment referencing the issue only on that particular substring.

braydonk:
Being able to add other error messages to filter kind of snowballed into a whole abstraction that we may not be interested in, open to opinions.

qingling128 (approved):
LGTM

PR #70, opened 2022-01-04
Update prometheus receivers

Mrod1598:
There was an issue in the CI on ops-agent that was causing the build to fail due to missing build tags. The build tags were added, so the receiver now needs to be updated.


qingling128 (approved):
LGTM

PR #71, opened 2022-01-05
Remove `--add-instance-id` flag

Mrod1598:
Remove use of flag that was removed in update to collector. [applicable changelog](https://github.com/open-telemetry/opentelemetry-collector/commit/e080cae46f19a3b30135db1dbe1070d676b1c2ae). Quoting:

```
- `--add-instance-id` is no longer operative; an instance ID will always be added.
```


qingling128 (approved):
LGTM

jsuereth:
QQ - if instance-id is always added, and it was previously set to false, what happens now?  There was a reason we set it to false, I'm guessing.

PR #77, opened 2022-01-21
Add fileexporter

sophieyfang:
Adding the fileexporter into the binaries will help us have some visibility on the metrics in OTLP format before it sends out

Context:
b/215605475
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/collector/integrationtests.md#adding-new-tests


qingling128 (approved):
LGTM modulo the conflicts are resolved.

PR #78, opened 2022-01-23
Added zookeeper component

cpheps:
This PR adds the Zookeeper metrics receiver to the operations collector.

PR #79, opened 2022-01-25
Add rabbit mongo zookeeper

cpheps:
Combined PR #78 (Zookeeper) and #76 (MongoDB) as well as RabbitMQ into a single PR. Also, updated a few of the dependencies due to OTel adding a new argument to the `config.ToClient`.


qingling128 (approved):
LGTM

PR #80, opened 2022-01-26
RFC: Add a new processor to convert gauges to cumulative sums.

igorpeshansky:
This can also handle non-cumulative sums.

This is an RFC to fix the limitation of [`windowsperfcountersreceiver`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/windowsperfcountersreceiver/windowsperfcounters_scraper.go#L120) that it can only create gauge metrics. Certain values gathered from the performance counters are cumulative (e.g., [those that start with "Total"](https://docs.microsoft.com/en-us/previous-versions//aa394298(v=vs.85))), so they need to be converted to sums. Those sums can later be processed by `normalizesumsprocessor`.


processor/normalizesumsprocessor/processor.go
```diff
@@ -77,11 +74,38 @@ func (nsp *NormalizeSumsProcessor) transformMetrics(rms pdata.ResourceMetrics) {
...
 func (nsp *NormalizeSumsProcessor) processMetric(resource pdata.Resource, metric pdata.Metric) bool {
-	dps := metric.Sum().DataPoints()
+	if metric.DataType() == pdata.MetricDataTypeSum && metric.Sum().IsMonotonic() {
+		return nsp.processDataPoints(resource, metric, metric.Sum().DataPoints())
+	}
+	if metric.DataType() == pdata.MetricDataTypeGauge && sliceContains(nsp.includeGauges, metric.Name()) {
+		newMetric := pdata.NewMetric()
+		metric.CopyTo(newMetric)
+		newMetric.SetDataType(pdata.MetricDataTypeSum)
+		newMetric.Sum().SetIsMonotonic(true)
```
aabmass, line 96: Is there a processor we could just stick in front of this one to convert the targeted Gauges to Sums?

quentinmit: +1, this doesn't seem like behavior that should be inside `normalizesumsprocessor`.

igorpeshansky: Added a `casttosum` processor (though it could probably be called `casttocumulative`). PTAL.

processor/casttosumprocessor/processor.go
```diff
@@ -0,0 +1,92 @@
...
+		return
+	}
+	if metric.DataType() == pdata.MetricDataTypeGauge {
+		newMetric := pdata.NewMetric()
+		metric.CopyTo(newMetric)
+		newMetric.SetDataType(pdata.MetricDataTypeSum)
+		metric.Gauge().DataPoints().CopyTo(newMetric.Sum().DataPoints())
+		newMetric.CopyTo(metric)
+	} else if metric.DataType() != pdata.MetricDataTypeSum {
+		ctsp.logger.Info(fmt.Sprintf("Configured metric %s is neither gauge nor sum\n", metric.Name()))
```
aabmass, line 88: nit could use the logger's formatting/structured logging

processor/casttosumprocessor/README.md
```diff
@@ -0,0 +1,27 @@
...
+The cast to sum processor (`casttosumprocessor`) converts (primarily gauge)
+metrics to cumulative sum metrics.
+
+Each value remains unchanged, but is transformed to a cumulative sum.
+
+## Configuration
+
+Configuration is specified through a list of metrics. The processor uses metric
+names to identify a set of input metrics and casts each of them to a cumulative
+sum.
```
aabmass, line 16:
```suggestion
monotonic sum.
```

processor/casttosumprocessor/processor.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (ctsp *CastToSumProcessor) transformMetrics(rms pdata.ResourceMetrics) {
+	ilms := rms.InstrumentationLibraryMetrics()
+	for j := 0; j < ilms.Len(); j++ {
+		ilm := ilms.At(j).Metrics()
+		newSlice := pdata.NewMetricSlice()
+		for k := 0; k < ilm.Len(); k++ {
+			metric := ilm.At(k)
+			ctsp.processMetric(rms.Resource(), metric)
```
aabmass, line 58: Im probably missing something here, but if this function modifies the point in place, why do you need the `newSlice.CopyTo(ilm)`?

aabmass (approved):
LGTM

igorpeshansky (commented):
Thanks for the review. PTAL.

processor/casttosumprocessor/config.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"go.opentelemetry.io/collector/config"
+)
+
+// Config defines the configuration for the processor.
+type Config struct {
+	*config.ProcessorSettings `mapstructure:"-"`
+
+	// List of input metrics to convert to sums
+	Metrics []string `mapstructure:"metrics"`
```
quentinmit, line 28: Name it `cast_to_sum_metrics` so we can later add other forms of casting if they come up?

processor/casttosumprocessor/README.md
```diff
@@ -0,0 +1,27 @@
+# Cast To Sum Processor
```
quentinmit, line 1: Consider putting this inside `agentmetricsprocessor`, which is already our catch-all processor for magic/bespoke processor behavior. It would reduce the amount of boilerplate you have here.

quentinmit (approved):

processor/casttosumprocessor/config.go
igorpeshansky: We could also create multiple instances of this processor with different options, right? Then each of them would have the `metrics` parameter…

I should probably change the above comment to just `// List of input metrics`.

processor/casttosumprocessor/README.md
igorpeshansky: Isn't `agentmetricsprocessor` designed to operate on the metrics produced by `hostmetricsreceiver`? It seems a bit hacky to also apply it to the IIS metrics (which is what this is designed for). Besides, having a separate processor makes it a lot easier to just blow away when the upstream receiver produces the correct metrics natively.

PR #82, opened 2022-01-28
cmd/otelopscol: remove error filter

braydonk:
This was only useful when the upstream fix didn't exist. The upstream
fixed is now much more desirable than this one. This reverts #68


qingling128 (approved):
LGTM

PR #83, opened 2022-02-04
Add a temporary googlecloud exporter that uses the new pdata exporter library

dashpole:
This is essentially the same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/7177, which is waiting for pdata to be made stable.  This will allow using the newly rewritten exporter in the meantime.  This is just a thin wrapper around the exporter libraries, which adds exporter-helper and implements the collector factory method.

This change is feature-gated.  To use the new exporter, run the collector with `--feature-gates=exporter.googlecloud.OTLPDirect`.  It is disabled by default in this PR.


punya (approved):

jsuereth (approved):

qingling128 (commented):
We are doing some parity tests on this to make sure there are no obvious regressions.

qingling128:
Oh NVM. I just saw this `This change is feature-gated. To use the new exporter, run the collector with --feature-gates=exporter.googlecloud.OTLPDirect. It is disabled by default in this PR.`

If that's the case, we can get it merge first. We'll do more validations before turning on that feature flag.

qingling128 (approved):
LGTM

PR #84, opened 2022-02-09
Updates opentelemetry-collector-contrib dependencies to v0.44.0

StefanKurek:
Updating the dependencies because wildfly needs to make use of a additional_jars config option in v0.44.0


qingling128 (approved):
LGTM

PR #88, opened 2022-03-24
Add varnish, iis, and sqlserver metrics receiver

JonathanWamsley:
This PR adds the Varnish, iis, and sql server metrics receiver to the operation collector.


cmd/otelopscol/components.go
```diff
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"github.com/observiq/observiq-otel-collector/receiver/varnishreceiver"
```
qingling128, line 18: Is this intentional or typo?

I vaguely recall there were some security concerns from the Otel community about how certain receiver works (especially the ones that calls an executable). Is this a workaround for that? Is it expected to be a temporary workaround and we plan to upstream it in the near future?

dehaansa: This is the workaround for the opentelemetry project's desire to not support any secondary executables as part of receivers. We've discussed it in recent meetings which I believe you were a part of, this is the plan, with no intention to upstream it unless the opentelemetry collector project changes its approach to secondary executables.

qingling128: Sync'd offline, we're gonna move the varnishreceiver to GoogleCloudPlatform/opentelemetry-operations-collector to avoid potential frictions with license / dependency policies. It will be addressed in a separate PR.

PR #89, opened 2022-03-28
Create and use Zap core to combat log spam

braydonk:
`scrapercontroller.go` from the OpenTelemetry Collector logs all scraper errors at Error level, causing a lot of log spam even from default installations. This PR implements a new Zap core that will change log levels before writing, either unilaterally or under some condition. We can use this on configuring the CollectorSettings to change all Error level logs from `scrapercontroller.go` to be Debug level.

### DIFF
--- a/cmd/otelopscol/main.go
+++ b/cmd/otelopscol/main.go
@@ -20,8 +20,11 @@ import (
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/service"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-collector/internal/env"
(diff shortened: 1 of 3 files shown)


cmd/otelopscol/main.go
```diff
@@ -44,6 +47,12 @@ func main() {
 	params := service.CollectorSettings{
 		Factories: factories,
 		BuildInfo: info,
+		LoggingOptions: []zap.Option{
+			levelchanger.NewLevelChangerOption(
+				zapcore.ErrorLevel,
+				zapcore.DebugLevel,
+				levelchanger.FilePathLevelChangeCondition("scrapercontroller.go")),
```
hsmatulisgoogle, line 54: Can you please point to the following line and say something like: this is the only error that gets mutated
https://github.com/open-telemetry/opentelemetry-collector/blob/831373ae6c6959f6c9258ac585a2ec0ab19a074f/receiver/scraperhelper/scrapercontroller.go#L198

PR #93, opened 2022-04-18
Add missing counter to IIS

Mrod1598:
Therer was a counter Path that was incorrect for a metric that has been fixed.


qingling128:
Could you elaborate a bit on the details for which metrics is changed?

Just wanna double check this does not introduce any backward incompatible changes.

Mrod1598:
> Could you elaborate a bit on the details for which metrics is changed?
> 
> Just wanna double check this does not introduce any backward incompatible changes.

Essentially, there was an issue with some of the paths from the last update to operations collector that wasn't caught due to github actions not running for IIS on collector-contrib. The metric wasn't correctly defined [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9358/files#diff-f7357a65213f8b912c7e3ac21e1f8763679cbff4b1aedb9bbaa88f49ccda6366L161). And instances weren't defined for another metric. It doesn't actually alter the v1 of IIS.

qingling128 (approved):
LGTM

PR #98, opened 2022-05-18
Add transform processor for moving resource attributes

dehaansa:
This new(ly supporting metrics) [processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) will allow a configuration like the following, which enables us to move forward with things like SQL Server & SAP Hana receivers which have resource attributes that need to be moved down to metric attributes.

```
processors:
  transform/sqlserver:
    metrics:
      queries:
        - set(attributes["host"], resource.attributes["sqlserver.host"]
  transform/saphana:
    metrics:
      queries:
        - set(attributes["host"], resource.attributes["saphana.host"]
```


quentinmit (approved):
LGTM to add this, but I wonder if we should revisit the idea of a "host" metric label now that service.name/service.namespace/service.instance_id are on the table?

PR #101, opened 2022-06-08
Update opentelemetry-collector to 0.53.0

schmikei:
Updated Otel Dependencies to 0.53.0


schmikei:
Just as a heads up this newer version has breaking JMX receiver changes which will need to modify the build process of the ops agent. This is the first step in getting it into the ops agent successfully.

qingling128 (approved):
LGTM

PR #103, opened 2022-06-21
Update Transformprocessor

Mrod1598:
Update trasnformprocessor to bring in new functionality for converting Sum datatypes.

Related PR:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/11041


braydonk (approved):
Approving because of the 0.54.0 contrib release delay. Assuming we'll update to that when it's ready.

PR #113, opened 2022-08-22
Go 1.18 update

cpheps:
Updated to Go 1.18 as minimum Go version. Changed all `go get` in Dockerfile to `go install` with specific version as `go get` for tools is not supported in Go 1.18.


internal/tools/go.mod
```diff
@@ -1,12 +1,178 @@
...
-	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/client9/misspell v0.3.4
-	github.com/golangci/golangci-lint v1.42.0
+	github.com/golangci/golangci-lint v1.48.0
 	github.com/google/addlicense v1.0.0
 	github.com/google/googet v2.13.0+incompatible
 	github.com/pavius/impi v0.0.3
 )
+
+require (
```
cpheps, line 13: Ran a `go mod tidy` on this module. Not sure if that's something we care to do since it's tooling code.

braydonk: This is fine imo

braydonk (approved):
LGTM, thanks for doing this!

PR #114, opened 2022-08-22
Do not fail otel if useragent can't be retrieved

sophieyfang:
Test evidence: b/230670850#comment61


qingling128:
The description sounds like the agent might not be able to correctly report user agent. Is there a backup path that ensures reasonable user agent is still being reported?

sophieyfang:
> The description sounds like the agent might not be able to correctly report user agent. Is there a backup path that ensures reasonable user agent is still being reported?

Our ops agent doesn't use this env UserAgent. Ops Agent has the UserAgent logic in its own code. 

For this repo's agent, if any error occur such env UserAgent will be empty, this will not cause this repo's agent to fail

qingling128 (approved):
LGTM

PR #118, opened 2022-09-06
Add sql query receiver

dehaansa:
Support for oracle DB was added to this receiver, so we'll be implementing oracle DB support using it in the ops-agent.


braydonk (approved):
LGTM 
(what a crazy amount of packages this receiver needs lol)

PR #121, opened 2022-09-08
components: add googlemanagedprometheus as an exporter

ridwanmsharif:
This change adds google managed prometheus as an exporter. To be used by the `prometheus` receiver the Ops Agent will support.


qingling128 (approved):
LGTM

PR #133, opened 2022-10-19
Remove strange "indir1ct" comments from go.mod

martijnvans:
Not sure where these comments came from; looks like odd typos that "go mod tidy" decided to preserve.


braydonk (approved):
Never noticed these :thinking: fascinating

PR #135, opened 2022-10-25
Also run presubmits on Windows

martijnvans:
The main goal of this PR was originally to run `integrationtest/hostmetrics_test.go` on Windows. And it did require some changes to start passing, because not all metrics are exported on Windows. However, I also had to fix a few other things in order to get the other checks (`checklicense`, `lint`, etc) passing on Windows.

The `.gitattributes` file is to avoid git sneakily replacing `\n` with `\r\n` for files that it considers text, such as our .go source files. This led to hard-to-debug errors where gofmt was complaining that files were not properly formatted. The solution is simply to disable this git feature. See golang.org/issue/9281 for precedent.

oh yeah, and I got some test failures from `receivers/varnish/client_test.go` because `filepath.Join("/exec/dir", "varnishstat") turns into `\exec\dir\varnishstat` with backslashes on Windows. Not sure how to best address that. I gave it a stab with filepath.Clean() but i'm not very happy with that.


integrationtest/hostmetrics_test.go
```diff
@@ -120,7 +133,9 @@ func loadExpectedMetrics(t *testing.T, expectedMetricsPath string) map[string]Ex
 		if _, ok := result[expect.Name]; ok {
 			t.Fatalf("Found multiple ExpectedMetric entries with Name=%q", expect.Name)
 		}
-		result[expect.Name] = expect
+		if len(expect.OnlyOn) == 0 || sliceContains(expect.OnlyOn, runtime.GOOS) {
```
braydonk, line 136: If `only_on` isn't in the yaml block, is it `nil` or an empty slice? I thought it was `nil`, but since this works maybe that's not the case.

braydonk (approved):

integrationtest/hostmetrics_test.go
martijnvans: len(nil) is zero. There is an internal style rule that apis should treat nil and an empty slice the same: go/go-style/decisions#nil-slices, and len() and probably other parts of the standard library adhere to this rule.

PR #140, opened 2022-11-14
Updated to otel v0.64.0

cpheps:
Updated to OTel [v0.64.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.64.0)


go.mod
```diff
@@ -3,65 +3,74 @@ module github.com/GoogleCloudPlatform/opentelemetry-operations-collector
...
-	go.opentelemetry.io/collector/exporter/loggingexporter v0.63.1
-	go.opentelemetry.io/collector/exporter/otlpexporter v0.63.1
-	go.opentelemetry.io/collector/exporter/otlphttpexporter v0.63.1
-	go.opentelemetry.io/collector/pdata v0.63.1
-	go.opentelemetry.io/collector/semconv v0.63.1
+	go.opentelemetry.io/collector v0.64.1
+	go.opentelemetry.io/collector/exporter/loggingexporter v0.64.1
+	go.opentelemetry.io/collector/exporter/otlpexporter v0.64.1
+	go.opentelemetry.io/collector/exporter/otlphttpexporter v0.64.1
+	go.opentelemetry.io/collector/extension/ballastextension v0.64.1
```
braydonk, line 44: Were this, `zpagesextension`, and `memorylimiterprocessor` brought in for a reason or was this an automated thing? If it was on purpose, I think we need to add them to `components.go` right? (I haven't heard anything about our plans with these but in fairness it would make sense for us to bring these in)

cpheps: I will double check. That wasn't intentional.

braydonk: Very weird, no idea how these were in `components.go` without being in `go.mod` all this time. Well at least it's correct now!

braydonk (approved):

PR #144, opened 2022-12-28
Add GPU process metrics to NVML receiver

suffiank:
To be added as following agent metrics:
- agent.googleapis.com/gpu/processes/lifetime_utilization {gpu_number, model, uuid, pid}
- agent.googleapis.com/gpu/processes/max_bytes_used {gpu_number, model, uuid, pid}

fyi - I added process labels {process, command, commandline, owner} to be consistent with existing agent process metrics. We can discuss if there's any issue.

cc - @LujieDuan


receiver/nvmlreceiver/client.go
```diff
@@ -278,6 +319,99 @@ func (client *nvmlClient) collectDeviceMemoryInfo() []nvmlMetric {
...
+
+func (client *nvmlClient) setProcessMetadataLabels(metric processMetric) processMetric {
+	metricName := fmt.Sprintf("nvml.processes{pid=%d}.metadata", metric.processPid)
+
+	process, err := process.NewProcess(int32(metric.processPid))
+	if err != nil {
+		msg := fmt.Sprintf("Unable to obtain process handle for pid %d to query for metadata on '%v'", metric.processPid, err)
+		client.issueWarningForFailedQueryUptoThreshold(int(metric.gpuIndex), metricName, msg)
+		return metric
+	}
```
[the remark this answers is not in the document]
suffiank: Yeah, agreed that's better.

receiver/nvmlreceiver/client.go
```diff
@@ -278,6 +321,93 @@ func (client *nvmlClient) collectDeviceMemoryInfo() []nvmlMetric {
...
+
+			stats, ret := nvml.DeviceGetAccountingStats(device, uint32(pid))
+			if ret != nvml.SUCCESS {
+				msg := fmt.Sprintf("Unable to query pid %d account statistics on '%v", pid, nvml.ErrorString(ret))
+				client.issueWarningForFailedQueryUptoThreshold(gpuIndex, metricName, msg)
+				continue
+			}
+
+			startTime := time.UnixMicro(int64(stats.StartTime))
+			if stats.IsRunning != 1 && startTime.Before(client.lastProcessInfoCollection) {
```
suffiank, line 350: @LujieDuan fyi

suffiank: I can try to confirm the behavior by running ephemeral jobs and seeing what it emits.

suffiank: image

Wdyt? Remove it? Might not leave a good impression. I say drop it and mention in the documentation that process needs to be alive at the time of capture.

suffiank: Though I haven't seen it in practice so far, this does raise the prospect of a race condition on which isRunning was set by NVIDIA and the process is dead by the time it's metadata is queried.

LujieDuan: Yes SGTM - let's drop those and have this documented. Thanks for testing this!

suffiank: Another concern this might raise is for a hypothetical inference scenario. A typical inference could run from 20-100 ms. If each inference request is served on a separate process (seems far-fetched, but maybe?) it could generate up to 1000 ephemeral processes. At that point NVIDIA's accounting statistics buffers would be full. And we'd end up reporting X metrics every collection, where X is NVIDIA's ring buffer size.

receiver/nvmlreceiver/client.go
```diff
@@ -278,6 +319,99 @@ func (client *nvmlClient) collectDeviceMemoryInfo() []nvmlMetric {
...
+				continue
+			}
+
+			metric := processMetric{
+				time:                   time.Now(),
+				processPid:             pid,
+				gpuIndex:               uint(gpuIndex),
+				runningTime:            time.Since(time.UnixMicro(int64(stats.StartTime))),
+				lifetimeGpuUtilization: uint64(stats.GpuUtilization),
+				lifetimeGpuMaxMemory:   stats.MaxMemoryUsage,
```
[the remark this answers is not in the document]
suffiank: Since it's possible to have some devices in MIG mode and some not, I changed the behavior to not disable process collection unless no devices support it.

receiver/nvmlreceiver/client.go
LujieDuan: Yeah that's a good point. I think it is a good idea to skip emitting deceased process metrics for now. So even if we are reporting the last X metrics, we have all the labels set correctly. 
For the race condition - I don't see a good way to avoid it if we choose to attach those process labels. We could document that and if it ever happens I believe the users can still group the metrics by PID when viewing the metrics.

LujieDuan (commented):
Thanks for the changes - this looks really good. Left a couple more comments:

receiver/nvmlreceiver/client.go
```diff
@@ -278,6 +347,91 @@ func (client *nvmlClient) collectDeviceMemoryInfo() []nvmlMetric {
...
 
+func (client *nvmlClient) collectProcessMetrics() []processMetric {
+	if client.disable || !client.collectProcessInfo {
+		return nil
+	}
+
+	processMetrics := make([]processMetric, 0)
+
+	for gpuIndex, device := range client.devices {
+		pids, ret := nvmlDeviceGetAccountingPids(device)
```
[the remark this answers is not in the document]
suffiank: Yeah, sounds fine to me. I can add that.

receiver/nvmlreceiver/client.go
suffiank: Sg to me.

suffiank:
@LujieDuan When you get a chance can you fix the CLA check? I think it's complaining after I committed one of your suggestions.

LujieDuan:
> @LujieDuan When you get a chance can you fix the CLA check? I think it's complaining after I committed one of your suggestions.

Done!

LujieDuan (approved):
LGTM!

suffiank:
@LujieDuan Great, thanks! Can you merge it? I don't have the permissions.

PR #147, opened 2023-01-17
Create a fork of the MongoDB receiver to support legacy systems

cpheps:
Created a fork of the MongoDB receiver at commit [2475cc7460a51924ecb3e80e40edc71fe0eaadc2](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/2475cc7460a51924ecb3e80e40edc71fe0eaadc2/receiver/mongodbreceiver) which was before any changes were made to the behavior for versions of MongoDB earlier than 4.0.

Changes that were made to original source:
- Removed the scraper and integration tests as they use helper structures from the `github.com/open-telemetry/opentelemetry-collector-contrib/internal/scrapertest` module and not accessible outside of the contrib repo.
- Removed the `go.mod` and `go.sum` files and integrated them into the main `go.mod` file for the repo.
- Updated to OTel v0.67.0 to match the rest of the repo.
  - Updated metadata.yaml to match new standards in v0.67.0


jefferbrecht:
LGTM 👍 I guess my only question is, is there anything relevant from the removed tests that need to be added to the Ops Agent tests downstream, or are we confident that we still have adequate coverage?

jefferbrecht (approved):

cpheps:
> is there anything relevant from the removed tests that need to be added to the Ops Agent tests downstream, or are we confident that we still have adequate coverage?

I think the fact that ops-agent has the kokoro integration test should suffice enough to replace the removed integration tests. It's not a perfect 1:1 replacement for all the removed test but covers the key functionality.

PR #148, opened 2023-01-18
Update building tags of the GPU receivers to fix the issue with Mac

LujieDuan:
The dependencies of the GPU receivers (go-nvml, go-dcgm) won't run properly on arm-based MacOS and prevent running presubmit tasks such as `make lint` locally when developing. Propose a change to explicitly exclude `darwin` using the build tag for the GPU receivers. 

Alternatively, we could exclude `arm64` instead of `darwin`; though that may require a slight change to the current makefile to have commands like `make lint` to work, since we are overriding `GOOS` and `GOARCH` in the makefile. 

CC @cpheps


cpheps:
I think these changes make sense. It's unlikely these GPU based receivers are applicable to makes so having `darwin` be exclude seems to be a safe choice.

receiver/dcgmreceiver/factory_others.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !windows
-// +build !windows
+//go:build !windows && !darwin
```
suffiank, line 15: Maybe we can just make this exclusive to Linux and change the corresponding file names to match?

LujieDuan: Good suggestion - done!

suffiank:
Lgtm :heavy_check_mark: I checked it out and run the unit tests.

suffiank (approved):

martijnvans (approved):

PR #149, opened 2023-01-19
Add null check for the nvml client in the shutdown function 

LujieDuan:
In the case when one of the receivers of the collector could not start, the shutdown function could be called without an instance of a nvml client. Add null check in the shutdown function. 
Screen Shot 2023-01-16 at 1 38 45 PM


suffiank (approved):
LGTM

martijnvans (approved):

PR #150, opened 2023-01-30
Add processor for controlling `instrumentation_scope`

quentinmit:
Will be used by Ops Agent to set the scope name and version for 3p integration receivers.


ridwanmsharif:
For context: Another way we could achieve the same behaviour is by using `metricstransform` to add the labels we want. We're going with this approach over using `metricstransform` to add the labels we care about because this approach will allow us to use one less exporter than otherwise in the Ops Agent.

A) The Ops Agent OTLP Receiver needs an exporter with:
- InstrumentationLibraryLabels flag turned on 
- ServiceResourceLabels flag turned on

B) The Ops Agent Agent Metrics receiver needs an exporter with:
- InstrumentationLibraryLabels flag turned off
- ServiceResourceLabels flag turned off

C) The Ops Agent 3P receiver metrics (if we add this `modifyscope`) processor needs an exporter with:
- InstrumentationLibraryLabels flag turned on (so we can actually get the instrumentation labels in the final metric)
- ServiceResourceLabels flag turned on

D) The Ops Agent 3P receiver metrics (if we use `metricstransform` processor and `add_label`) needs an exporter with:
- InstrumentationLibraryLabels flag turned off (since we add the labels directly to the metric in the Ops Agent) 
- ServiceResourceLabels flag turned on

Therefore since C and A use the same settings, this solution can use one less exporter.

processor/modifyscopeprocessor/factory.go
```diff
@@ -0,0 +1,71 @@
...
+		params,
+		cfg,
+		nextConsumer,
+		metricsProcessor.ProcessMetrics,
+		processorhelper.WithCapabilities(processorCapabilities))
+}
+
+// validateConfiguration validates the input configuration has all of the required fields for the processor
+// An error is returned if there are any invalid inputs.
+func validateConfiguration(config *Config) error {
```
ridwanmsharif, line 69: [optional] Based on how the Ops Agent is using this processor, are there some formats we want to validate against? Must have some specific prefix for the instrumentation_scope for example

ridwanmsharif (approved):

processor/modifyscopeprocessor/factory.go
quentinmit: Since this isn't for upstreaming or direct user use I don't particularly care if it's hard to use; I'd rather just keep this code as simple as possible here.

PR #156, opened 2023-03-16
Add gpu build tag to toggle gpu support

braydonk:
This PR adds a new `gpu` build tag to control building GPU support into the collector. These previously used platforms as the build constraints, only building on Linux. This build tag setup allows us to toggle GPU support on and off regardless to enable building the collector without CGO and to make future efforts like ARM support simpler.


receiver/dcgmreceiver/scraper_gpu_test.go
```diff
@@ -28,13 +28,14 @@ import (
...
 	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/receiver"
 	"go.uber.org/zap/zaptest"
 
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-collector/receiver/dcgmreceiver/testprofilepause"
 )
 
 func TestScrapeWithGpuPresent(t *testing.T) {
-	settings := componenttest.NewNopReceiverCreateSettings()
+	var settings receiver.CreateSettings
```
braydonk, line 38: `componenttest.NewNopReceiverCreateSettings()` was deprecated in collector v0.69.0

receiver/nvmlreceiver/testcudakernel/test_cuda_kernel.cc
```diff
@@ -16,30 +16,33 @@
 // https://stackoverflow.com/questions/27900849/unable-to-use-cublasxt
```
braydonk, line 16: Oops, I forgot I had my editor configured to run `clang-format` on save for C/C++ files. Shall I revert the formatting changes?

Makefile
```diff
@@ -90,10 +96,22 @@ misspell:
 build:
 	go build -o ./bin/$(OTELCOL_BINARY) $(LD_FLAGS) ./cmd/otelopscol
 
+.PHONY: build_gpu_support
+build_gpu_support:
+	go build -tags=gpu -o ./bin/$(OTELCOL_BINARY) $(LD_FLAGS) ./cmd/otelopscol
```
LujieDuan, line 101: I have a suggestion here: 
Instead of making targets for each combination of the tags, we could use a variable:
Remove `build_gpu_support` target and modify the original `build` target:
```
build:
	go build -tag=$(GO_TAGS) -o ./bin/$(OTELCOL_BINARY) $(LD_FLAGS) ./cmd/otelopscol
```
and when calling we can do `make build` for gpu off, and `make GO_TAGS=gpu` for gpu on. And we can do this for the testing targets as well. This may be beneficial if we have more tags, but how to run with `gpu` would become less obvious. 

I don't have a strong preference between those two, so I will ask you to decide!

receiver/nvmlreceiver/testcudakernel/test_cuda_kernel.cc
LujieDuan: I think we can keep this - no need to revert.

LujieDuan (approved):
LGTM!

Makefile
braydonk: Done in [f3be5f6](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/156/commits/f3be5f62d83b33c2a8aecef71ad73eb27211f5ef)

PR #158, opened 2023-03-21
Update to otel version v0.74.0

braydonk:
* Ran `make update-components`
* Update to go v1.20
* Manually updated version of mdatagen
* Regenerated metadata for new version
* Fixed new changes in metadata generation


braydonk:
That Kokoro check is guaranteed to fail. Until this gets merged to master to upload the updated container, it won't work.

receiver/varnishreceiver/internal/metadata/generated_metrics.go
```diff
@@ -32,7 +27,7 @@ func (ms *MetricSettings) Unmarshal(parser *confmap.Conf) error {
 	if err != nil {
 		return err
 	}
-	ms.enabledProvidedByUser = parser.IsSet("enabled")
+	ms.enabledSetByUser = parser.IsSet("enabled")
```
franciscovalentecastro, line 30: Why is this name `enabledProvidedByUser` update ?

receiver/varnishreceiver/scraper.go
```diff
@@ -41,7 +41,7 @@ func newVarnishScraper(settings receiver.CreateSettings, config *Config) *varnis
 	return &varnishScraper{
 		telemetrySettings: settings.TelemetrySettings,
 		config:            config,
-		mb:                metadata.NewMetricsBuilder(metadata.DefaultMetricsSettings(), settings),
+		mb:                metadata.NewMetricsBuilder(metadata.DefaultMetricsBuilderConfig(), settings),
```
franciscovalentecastro, line 44: Is this update needed ? 

I'm seeing the other receivers (`nvml`, `mongodb`, `dcgm`) use the `mbConfig.Metrics = s.config.Metrics` to set the metrics configuration, but this one is using the `Default`, so I understand this update creates the same thing as before.

braydonk: Because this used the `DefaultMetricsSettings`, I did a direct mapping to the new way of doing things. I don't actually know why this never used metrics from `Config`. I will raise this to ObservIQ, but that will be another fix PR outside of this one.

receiver/varnishreceiver/internal/metadata/generated_metrics.go
braydonk: It was changed in the generation templates: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/4524ffcac423dcd6743f012e3b24808e6346a7cd/cmd/mdatagen/templates/metrics.go.tmpl#L26
I don't really know why, I didn't pay much mind to it because it doesn't end up being used outside this generated package

franciscovalentecastro: Thanks!

receiver/varnishreceiver/scraper.go
franciscovalentecastro: Thanks! Yeah, now it makes sense.

franciscovalentecastro (approved):

PR #157, opened 2023-03-17
Removing unused packaging code and update README

braydonk:
This repo was originally designed for something completely different than it's used for today. It would have been incredibly misleading for anyone discovering the repo without any context. This PR is to completely gut all of the unused packaging code and make targets, as well as revamp the README entirely to reflect what this collector is actually used for today.

Closes #155

PR #161, opened 2023-03-30
Fix minor formatting error in upgrade-opentelemetry.md

igorpeshansky:
Following up on my botched suggestion in #157.


braydonk (approved):
Thanks, I forgot to check the Rich Diff after applying that suggestion.

PR #166, opened 2023-05-16
Upgrade docker/{docker,distribution}

jefferbrecht:
Dependabot-suggested upgrade.


braydonk (approved):
idk where we're even using this but that's unimportant right now, LGTM

PR #167, opened 2023-05-26
[WIP] processhandleprocessor: created process handle processor

braydonk:


braydonk:
Got this feature merged upstream which means this black magic can now be erased

PR #169, opened 2023-06-29
receiver/prometheusreceiver: fork prometheus receiver so we can support untyped metrics

ridwanmsharif:
This change forks the prometheus receiver from the contrib repo. It also removes the unused `prometheusexecreceiver` since it was unused.

I skip the linter for the new repo, so we can more easily keep parity with upstream for if/when we use upstream again.


.github/header-checker-lint.yml
```diff
@@ -4,6 +4,8 @@ ignoreFiles:
   - receiver/**/generated_metrics.go
   - receiver/**/generated_metrics_test.go
   - receiver/mongodbreceiver/**/*.go
+  - receiver/prometheusreceiver/**/*.go
```
braydonk, line 7: Doesn't this pattern cover both or nah?

go.mod
```diff
@@ -168,7 +174,7 @@ require (
 	github.com/form3tech-oss/jwt-go v3.2.5+incompatible // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
 	github.com/gabriel-vasile/mimetype v1.4.1 // indirect
-	github.com/go-kit/log v0.2.1 // indirect
+	github.com/go-kit/log v0.2.1
```
braydonk, line 177: This might need a `go mod tidy` if you haven't already, the directs and indirects should be in different places not intermingled. `go mod tidy` should fix that

ridwanmsharif: This is what i got after `go mod tidy`. I think this might've happened because I'm pinning the version of `go.opentelemetry.io/collector/featuregate` to  `v0.74.0`. The google cloud exporter relies on this version (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f08863c899dc4fd26342000a16658d06dc86e569/exporter/googlecloudexporter/go.mod#L12) and leaving it as an indirect dep broke things because the exporter relies on methods that aren't available in the newer versions.

.github/header-checker-lint.yml
ridwanmsharif: I think you're right. Just tested it, god I had globstar

ridwanmsharif:
Also for sanity: https://github.com/GoogleCloudPlatform/ops-agent/pull/1312 this is the Ops Agent PR with integration tests and 3P tests running that uses this branch.

go.mod
braydonk: I wish I understood why it isn't sorting the dependencies the way it should be. These becoming direct makes sense, but it's strange that they are intertwined with the indirect dependencies like this.

I won't block the PR on it, but something does seem weird there.

braydonk (approved):

PR #170, opened 2023-07-06
prometheusreceiver: add metric attribute for untyped metrics

ridwanmsharif:
This change adds a metric attribute that encodes whether the original promehteus metric was untyped. It is still convereted to an OTLP gauge but with an additional attribute so the exporters and processors can use this information to drop the metric to support it differently than gauges.

### DIFF
--- a/receiver/prometheusreceiver/internal/metricfamily.go
+++ b/receiver/prometheusreceiver/internal/metricfamily.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/prometheus/prometheus/model/exemplar"
 	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/model/textparse"
 	"github.com/prometheus/prometheus/model/value"
 	"github.com/prometheus/prometheus/scrape"
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -33,12 +34,14 @@ import (
 )
 
 const (
-	traceIDKey = "trace_id"
-	spanIDKey  = "span_id"
+	traceIDKey           = "trace_id"
+	spanIDKey            = "span_id"
+	PrometheusUntypedKey = "promethues_untyped_metric"
 )
 
 type metricFamily struct {
 	mtype pmetric.MetricType
+	pType textparse.MetricType
 	// isMonotonic only applies to sums
 	isMonotonic bool
 	groups      map[uint64]*metricGroup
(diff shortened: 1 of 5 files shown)


damemi (approved):
lgtm, @dashpole may want to take a look too

receiver/prometheusreceiver/metrics_receiver_helper_test.go
```diff
@@ -447,6 +447,9 @@ func compareAttributes(attributes map[string]string) numberPointComparator {
 		if req {
 			for k, v := range attributes {
 				val, ok := numberDataPoint.Attributes().Get(k)
+				if !ok {
```
braydonk, line 450: I think you can just give `require.True` a message instead of checking the false condition here. https://pkg.go.dev/github.com/stretchr/testify/require#True

receiver/prometheusreceiver/metrics_receiver_helper_test.go
```diff
@@ -447,6 +447,9 @@ func compareAttributes(attributes map[string]string) numberPointComparator {
...
+					log.Printf("Nothing found")
```
braydonk, line 451: I'd suggest using [t.Log](https://pkg.go.dev/testing#T.Log) for these sorts of things when in a test, that way the input is easier to follow in parallel tests and is controlled by the `-v` flag.
(doesn't matter if you use my other suggestion here)

braydonk (approved):
just silly test code nitpicks, otherwise LGTM

receiver/prometheusreceiver/metrics_receiver_helper_test.go
ridwanmsharif: ughghghghg my bad added this here so I could test using breakpoints, forgot to remove this

dashpole (approved):

receiver/prometheusreceiver/internal/metricfamily.go
```diff
@@ -33,12 +34,14 @@ import (
 )
 
 const (
-	traceIDKey = "trace_id"
-	spanIDKey  = "span_id"
+	traceIDKey           = "trace_id"
+	spanIDKey            = "span_id"
+	PrometheusUntypedKey = "promethues_untyped_metric"
```
damemi, line 39: @ridwanmsharif sorry but I just noticed this typo: `promethues_untyped_metric` -> `prometheus_untyped_metric`

ridwanmsharif: good catch 🤦

ridwanmsharif: Sent over https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/172

PR #174, opened 2023-07-18
DCGM receiver: Soft disable the receiver when DCGM not installed

LujieDuan:
b/279022165

Modify the start up behaviour of the DCGM receiver to have similar logic as the NVML receiver - when the receiver is enabled and the DCGM app is not installed OR not running, soft disable the receiver, emit no metrics, and log the warning message. Throwing errors in this case could cause the entire collector to exit, which is normally not the expected behaviour. 

Tests in the Ops Agent: https://github.com/GoogleCloudPlatform/ops-agent/pull/1338

### DIFF
--- a/receiver/dcgmreceiver/client.go
+++ b/receiver/dcgmreceiver/client.go
@@ -31,6 +31,7 @@ const maxWarningsForFailedDeviceMetricQuery = 5
 
 type dcgmClient struct {
 	logger                         *zap.SugaredLogger
+	disable                        bool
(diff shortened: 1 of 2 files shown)


LujieDuan:
cc @braydonk

suffiank (approved):
Lgtm :)

LujieDuan:
Hey @braydonk I need a approval with write access :) thanks!

braydonk (approved):
Did not realize I had to give this too. LGTM!

PR #178, opened 2023-07-25
DCGM: Enable the receiver to retry dcgm initialization at each collection time

LujieDuan:
[b/292514145](http://b/292514145)

- First, revert #174;
- Modify the receiver to reinitialize the connection to DCGM by trying to create a new client at each collection time;
- The DCGM lib has an issue right now, where dcgm.init() can't be called more than once when DCGM is not installed: https://github.com/NVIDIA/go-dcgm/issues/13. Added a workaround by checking the `libdcgm.so` directly before calling dcgm.init().
- Testing in the Ops Agent: https://github.com/GoogleCloudPlatform/ops-agent/pull/1348

### DIFF
--- a/receiver/dcgmreceiver/client.go
+++ b/receiver/dcgmreceiver/client.go
@@ -17,16 +17,10 @@
 
 package dcgmreceiver
 
-/*
-#include <stdlib.h>
-#include <dlfcn.h>
-*/
-import "C"
 import (
 	"errors"
 	"fmt"
 	"time"
-	"unsafe"
 
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
 	"go.opentelemetry.io/collector/receiver/scrapererror"
@@ -108,33 +102,18 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
 	}, nil
 }
 
-// isDcgmInstalled returns true if it was able to load the DCGM shared library.
-// TODO(b/293585142): Remove this function and use dcgm.Init() to check if DCGM
-// is installed after https://github.com/NVIDIA/go-dcgm/issues/13 is fixed
-func isDcgmInstalled() bool {
-	const (
-		dcgmLib = "libdcgm.so"
-	)
-	lib := C.CString(dcgmLib)
-	defer C.free(unsafe.Pointer(lib))
-
-	dcgmLibHandle := C.dlopen(lib, C.RTLD_LAZY|C.RTLD_GLOBAL)
-	return dcgmLibHandle != nil
-}
-
 // initializeDcgm tries to initialize a DCGM connection; returns a cleanup func
 // only if the connection is initialized successfully without error
 func initializeDcgm(config *Config, logger *zap.Logger) (func(), error) {
-	if !isDcgmInstalled() {
-		return nil, fmt.Errorf("cannot initialize a DCGM client; DCGM is not installed")
-	}
 	isSocket := "0"
 	dcgmCleanup, err := dcgmInit(config.TCPAddr.Endpoint, isSocket)
 	if err != nil {
+		msg := fmt.Sprintf("Unable to connect to DCGM daemon at %s on %v; Is the DCGM daemon running?", config.TCPAddr.Endpoint, err)
+		logger.Sugar().Warn(msg)
 		if dcgmCleanup != nil {
 			dcgmCleanup()
 		}
-		return nil, fmt.Errorf("Unable to connect to DCGM daemon at %s on %v; Is the DCGM daemon running?", config.TCPAddr.Endpoint, err)
+		return nil, fmt.Errorf("%s", msg)
 	}
 	logger.Sugar().Infof("Connected to DCGM daemon at %s", config.TCPAddr.Endpoint)
 	return dcgmCleanup, nil
--- a/receiver/dcgmreceiver/client.go
+++ b/receiver/dcgmreceiver/client.go
@@ -17,10 +17,16 @@
 
 package dcgmreceiver
 
+/*
+#include <stdlib.h>
+#include <dlfcn.h>
+*/
+import "C"
 import (
 	"errors"
 	"fmt"
 	"time"
+	"unsafe"
 
 	"github.com/NVIDIA/go-dcgm/pkg/dcgm"
 	"go.opentelemetry.io/collector/receiver/scrapererror"
@@ -75,6 +81,19 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
 	for _, f := range unavailableFields {
 		logger.Sugar().Warnf("Field '%s' is not supported. Metric '%s' will not be collected", dcgmIDToName[f], dcgmNameToMetricName[dcgmIDToName[f]])
 	}
+	// If there is no metrics to watch after filtering the user configuration
+	// with device supported fields, return an client with an empty device list
+	if len(enabledFields) == 0 {
+		return &dcgmClient{
+			logger:                         logger.Sugar(),
+			enabledFieldIDs:                enabledFields,
+			enabledFieldGroup:              dcgm.FieldHandle{},
+			deviceIndices:                  make([]uint, 0),
+			devicesModelName:               make([]string, 0),
+			devicesUUID:                    make([]string, 0),
+			deviceMetricToFailedQueryCount: make(map[string]uint64, 0),
+		}, nil
+	}
 	enabledFieldGroup, err := setWatchesOnEnabledFields(config, logger, deviceGroup, enabledFields)
 	if err != nil {
 		return nil, fmt.Errorf("Unable to set field watches on %w", err)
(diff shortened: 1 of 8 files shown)


jefferbrecht:
@suffiank are you able to take a look at this?

suffiank (approved):
Lgtm :) cc @jefferbrecht

receiver/dcgmreceiver/client.go
```diff
@@ -83,14 +81,26 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
...
+	if len(enabledFields) == 0 {
+		return &dcgmClient{
+			logger:                         logger.Sugar(),
+			enabledFieldIDs:                enabledFields,
+			enabledFieldGroup:              dcgm.FieldHandle{},
+			deviceIndices:                  make([]uint, 0),
+			devicesModelName:               make([]string, 0),
+			devicesUUID:                    make([]string, 0),
+			deviceMetricToFailedQueryCount: make(map[string]uint64, 0),
+		}, nil
```
igorpeshansky, line 95: This seems like code duplication, and a bunch of unnecessary work to discover the available devices, etc, only to ignore them here. Unless I'm missing something, both `requestedFieldIDs` and `supportedFieldIDs` don't depend on any of the other computed values, and thus can be queried first, so this could be short-circuited way earlier.

To be clear, here's what I'm suggesting:
```
func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
	deviceIndices := make([]uint, 0)
	names := make([]string, 0)
	UUIDs := make([]string, 0)
	enabledFieldGroup := dcgm.FieldHandle{},
	requestedFieldIDs := discoverRequestedFieldIDs(config)
	supportedFieldIDs, err := getAllSupportedFields()
	…
	enabledFields, unavailableFields := filterSupportedFields(requestedFieldIDs, supportedFieldIDs)
	…
	if len(enabledFields) != 0 {
		deviceIndices, names, UUIDs, err = discoverDevices(logger)
		…
		deviceGroup, err = createDeviceGroup(logger, deviceIndices)
		…
		enabledFieldGroup, err = setWatchesOnEnabledFields(config, logger, deviceGroup, enabledFields)
		…
	}
	return &dcgmClient{
		logger:                         logger.Sugar(),
		enabledFieldIDs:                enabledFields,
		enabledFieldGroup:              enabledFieldGroup,
		deviceIndices:                  deviceIndices,
		devicesModelName:               names,
		devicesUUID:                    UUIDs,
		deviceMetricToFailedQueryCount: make(map[string]uint64),
	}, nil
}
```

receiver/dcgmreceiver/client.go
```diff
@@ -100,12 +110,28 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
 	}, nil
 }
 
+// TODO: b/293585142 - Remove this function and use dcgm.Init() to check if DCGM is
```
igorpeshansky, line 113:
```suggestion
// TODO(b/293585142): Remove this function and use dcgm.Init() to check if DCGM is
```

receiver/dcgmreceiver/client.go
```diff
@@ -100,12 +110,28 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
...
+// installed after https://github.com/NVIDIA/go-dcgm/issues/13 is fixed
+func isDcgmInstalled() bool {
```
igorpeshansky, line 115: I'd add a comment explaining what this function does, e.g.:
```
// isDcgmInstalled returns true if it was able to load the dcgm shared library.
// TODO(…): …
func isDcgmInstalled() bool {
	…
}
```

receiver/dcgmreceiver/scraper.go
```diff
@@ -31,23 +31,45 @@ import (
...
+// initClient will try to create a new dcgmClient if currently has no client;
+// it will try to initialize the communication with the DCGM service; if
+// success, create a client; only return errors if DCGM service is available but
+// failed to create client.
+func (s *dcgmScraper) initClient() error {
+	if s.client != nil {
+		return nil
+	}
+	if !isDcgmInstalled() {
+		s.settings.Logger.Warn("can not initialize a DCGM client; DCGM is not installed.")
```
igorpeshansky, line 54: Nit: cannot.
```suggestion
		s.settings.Logger.Warn("cannot initialize a DCGM client; DCGM is not installed.")
```
Also in the comment below.

receiver/dcgmreceiver/scraper.go
```diff
@@ -31,23 +31,45 @@ import (
...
+		// If can not connect to DCGM, return no error and retry at next
+		// collection time
+		return nil
+	}
+	client, err := newClient(s.config, s.settings.Logger)
 	if err != nil {
 		return err
 	}
+	s.client = client
+	s.handleCleanup = dcgmCleanup
```
igorpeshansky, line 68: I'm honestly a bit confused why you can no longer store `handleCleanup` in the client. Is this because `newClient` may be called more than once? But then so would `initializeDcgm`…

receiver/dcgmreceiver/client.go
LujieDuan: Done! Much cleaner right now. Thanks!

receiver/dcgmreceiver/client.go
LujieDuan: Done.

receiver/dcgmreceiver/scraper.go
LujieDuan: Done.

receiver/dcgmreceiver/scraper.go
LujieDuan: Done! Moved the `handleCleanup` back to the client. 

Previously, I moved `handleCleanup` up to the scraper as I wanted to call `initializeDcgm()` in the scraper to test if the dcgm can be initialized before create a client. But I agree it can be confusing. 

Right now I've moved the `handleCleanup` back to the client. When calling `newClient()` from the scraper, in order to tell if the error is an initialization error (which can be ignored and retried) or other error, I've defined an error type `ErrDcgmInitialization` and check that.

receiver/dcgmreceiver/client.go
LujieDuan: Done.

receiver/dcgmreceiver/client.go
```diff
@@ -60,47 +63,46 @@ var dcgmInit = func(args ...string) (func(), error) {
...
-	enabledFieldGroup, err := setWatchesOnEnabledFields(config, logger, deviceGroup, enabledFields)
-	if err != nil {
-		return nil, fmt.Errorf("Unable to set field watches on %w", err)
+
+	if len(enabledFields) != 0 {
+		deviceIndices, names, UUIDs, err = discoverDevices(logger)
+		if err != nil {
+			return nil, err
+		}
+
```
igorpeshansky, line 92: We should be consistent about blank lines between these conditionals — I'd remove this one or add one after line 96. Also in line 81.

receiver/dcgmreceiver/client.go
```diff
@@ -110,7 +112,8 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
 	}, nil
 }
 
-// TODO: b/293585142 - Remove this function and use dcgm.Init() to check if DCGM is
+// isDcgmInstalled returns true if it was able to load the dcgm shared library.
```
igorpeshansky, line 115: [Optional] Isn't DCGM an acronym? If so, it should be capitalized in comments here and below.

receiver/dcgmreceiver/scraper.go
```diff
@@ -41,13 +42,33 @@ func newDcgmScraper(config *Config, settings receiver.CreateSettings) *dcgmScrap
...
+// success, create a client; only return errors if DCGM service is available but
+// failed to create client.
+func (s *dcgmScraper) initClient() error {
+	if s.client != nil {
+		return nil
+	}
+	if !isDcgmInstalled() {
+		s.settings.Logger.Warn("cannot initialize a DCGM client; DCGM is not installed.")
+		return nil
+	}
```
igorpeshansky, line 56: It seems like you could move this block into `newClient` (or even, optionally, `initializeDcgm`) and have it return `ErrDcgmInitialization`. The behavior of the scraper would be identical, but the code would become simpler (callers wouldn't need to know about the separate "installed" check).

receiver/dcgmreceiver/client.go
```diff
@@ -35,8 +35,11 @@ import (
 
 const maxWarningsForFailedDeviceMetricQuery = 5
 
+var ErrDcgmInitialization = errors.New("error initialize dcgm")
```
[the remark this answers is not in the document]
LujieDuan: Done!

receiver/dcgmreceiver/client.go
LujieDuan: Done!

receiver/dcgmreceiver/client.go
LujieDuan: Done!

receiver/dcgmreceiver/scraper.go
LujieDuan: Good point! Moved this check to `initializeDcgm()` and clean up the error message a bit, and the logic feels much clearer! Thanks!

receiver/dcgmreceiver/client_test.go
```diff
@@ -12,42 +12,23 @@
...
 func TestNewDcgmClientOnInitializationError(t *testing.T) {
-	realDcgmInit := dcgmInit
-	defer func() { dcgmInit = realDcgmInit }()
-	dcgmInit = func(args ...string) (func(), error) {
-		return nil, fmt.Errorf("No DCGM client library *OR* No DCGM connection")
-	}
-
-	seenDcgmConnectionWarning := false
-	logger := zaptest.NewLogger(t, zaptest.WrapOptions(zap.Hooks(func(e zapcore.Entry) error {
-		if e.Level == zap.WarnLevel && strings.Contains(e.Message, "Unable to connect to DCGM daemon") {
```
igorpeshansky, line 41: I'm not sure we wanted to remove the test for this failure mode. There are two ways creating a new client could fail, right? Or am I missing something?

LujieDuan: Good point! Actually it looks https://github.com/NVIDIA/go-dcgm/issues/13 just got fixed. With this fix, we should not need to check the `.so` file in our code anymore. Let me upgrade and remove the temporary `isDcgmInstalled()` function.

receiver/dcgmreceiver/client_gpu_test.go
```diff
@@ -53,7 +53,7 @@ type modelSupportedFields struct {
 func TestSupportedFieldsWithGolden(t *testing.T) {
 	config := createDefaultConfig().(*Config)
 	client, err := newClient(config, zaptest.NewLogger(t))
-	require.Nil(t, err)
+	require.Nil(t, err, "can not initialize DCGM. Install and run DCGM before running tests.")
```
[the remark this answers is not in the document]
LujieDuan: Done!

igorpeshansky (approved):
LGTM :shipit:

PR #185, opened 2023-10-17
Upgrade to OpenTelemetry v0.87.0

braydonk:
This PR updates all OpenTelemetry components to their latest version.

`googlemanagedprometheusexporter` changes are related to: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/26071
`prometheusreceiver` changes are related to: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24256
`aggregation_temporality` changes are related to: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24808


ridwanmsharif (approved):

exporter/googlemanagedprometheusexporter/config.go
```diff
@@ -41,8 +40,9 @@ type GMPConfig struct {
 type MetricConfig struct {
 	// Prefix configures the prefix of metrics sent to GoogleManagedPrometheus.  Defaults to prometheus.googleapis.com.
 	// Changing this prefix is not recommended, as it may cause metrics to not be queryable with promql in the Cloud Monitoring UI.
-	Prefix       string                 `mapstructure:"prefix"`
-	ClientConfig collector.ClientConfig `mapstructure:",squash"`
+	Prefix       string                         `mapstructure:"prefix"`
+	ClientConfig collector.ClientConfig         `mapstructure:",squash"`
+	Config       googlemanagedprometheus.Config `mapstructure:",squash"`
```
ridwanmsharif, line 45: Okay I've been thinking about this and I find this name to be pretty difficult to reason about. WDYT of `GMPConversionConfig` instead?

This config controls 4 things:
- Adding target_info as a metric (otel resource attributes as labels)
- Adding otel_scope_info as a metric (otel instrumentation scope attributes as labels)
- Adding untyped metrics (kind of a misnomer, it doesn't add untyped metrics, instead it adds the counter half of the untyped metrics)
- Add metric suffixes (this is also a bit of a misnomer, this controls normalization of prometheus metrics. There are a lot of things that normalization does and https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/21743 covers some examples of how this breaks users. I'm bringing this up because this refactor enables this option by default. The `pkg.translator.prometheus.NormalizeName` feature gate is again turned off by default so nothing *should* change with the refactor but we should confirm). I will say, for this one specifically I'm not sure if our prometheus test cases are enough to catch it. Not sure if we test both the metrics with existing suffixes and without suffixes.

This is all configurable from the exporter side now so I'm not worried about blocking this PR other than the naming though.

When we do pull this into the Ops Agent though, we need to figure out a few things:
- What configuration settings to use for the extra metrics target info and scope info settings, for both the prometheus and otlp pipelines
- Whether to make those settings configurable in the ops agent (I think this is unecesary tbh, the answer is probably no but figured I'd ask)
- Whether the new defaults for `AddMetricsSuffix` cause breaking changes for us. I don't think it will given the `pkg.translator.prometheus.NormalizeName` is disabled but the OTel releases flipflopped this setting and then changed the translator again so we should confirm this (bonus points for beefing up the Ops Agent integration tests to catch this)

braydonk: Discussed offline, while I agree the name is better we want to stick closer to upstream where we can manage it to make the manual diffing between the two as easy as we can manage. (Hopefully we just unfork soon then it won't matter but until then :sad: )

PR #197, opened 2024-01-19
change featuregate name

braydonk:
The featuregate name changed in the googlemanagedprometheus exporter library in `opentelemetry-operations-go` library. This PR changes the featuregate name to match.

I would add a test for this, but I am trying to reduce how much we deviate from upstream to make it easier to migrate for future changes we need to move forward in our fork.

### DIFF
--- a/.build/Dockerfile
+++ b/.build/Dockerfile
@@ -2,7 +2,10 @@ FROM golang:1.21.6-bullseye
 
 RUN apt-get update && apt-get install gettext-base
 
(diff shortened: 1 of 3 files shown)


jefferbrecht (approved):
LGTM, the test failure is almost certainly unrelated to this change, it is weird though because Quentin's PR from the other day was fine. Maybe it's just flaky, I'll try re-running it.

braydonk:
We've decided to merge despite the borked kokoro issues which are occurring due to an issue with running `golangci-lint` in the new golang docker container.

PR #198, opened 2024-01-19
Rename feature gate in the `googlemanagedprometheus` exporter

quentinmit:
`opentelemetry-operations-go v0.45.0` renames the `gcp.untyped_double_export` feature gate to `gcp.untypedDoubleExport` and this wasn't caught by any of the tests in this repo. :(

On the bright side, the Ops Agent integration tests did catch it.


quentinmit:
Apparently #197 was also in flight with the same fix :)

PR #209, opened 2024-05-23
service: add levelchanger for localhostgate warn

braydonk:
This PR adds a new levelchanger option that will adjust the localhost default featuregate warning to debug. We don't want to turn on that featuregate, so we have to do this to get rid of the log.


LujieDuan (approved):
LGTM!

PR #219, opened 2024-07-16
Implement an initial set of uniform DCGM GPU metrics in `dcgmreceiver`.

igorpeshansky:
This PR implements an initial set of DCGM GPU metrics from the uniform set defined in [go/gce-gke-gpu-dcgm-metrics](http://go/gce-gke-gpu-dcgm-metrics) in `dcgmreceiver`.

In the process, it changes the GPU device-identifying metric attributes into resource attributes, adds support for older (and newer) GPU devices, and removes the assumption of 1:1 metric/DCGM field correspondence.


receiver/dcgmreceiver/client.go
```diff
@@ -66,15 +70,21 @@ func newClient(config *Config, logger *zap.Logger) (*dcgmClient, error) {
 	UUIDs := make([]string, 0)
 	enabledFieldGroup := dcgm.FieldHandle{}
 	requestedFieldIDs := discoverRequestedFieldIDs(config)
-	supportedFieldIDs, err := getAllSupportedFields()
+	supportedRegularFieldIDs, err := getSupportedRegularFields(requestedFieldIDs, logger)
+	if err != nil {
+		// TODO: If there is error querying the supported fields at all, let the
```
jefferbrecht, line 75: What's the AI for the TODO? It's not clear whether the rest of the comment is describing the current behaviour or a desired future behaviour.

igorpeshansky: This TODO was actually a reminder for me to revisit this comment, which simply copied the one in lines 81–82. However, your comment made me realize that, unlike in the case of profiling metrics, a failure here would be catastrophic, and should not allow a client to be created. Fixed in https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/219/commits/d0ad1ad8438c58a79e19e4c9eaac0d652eee5043.

receiver/dcgmreceiver/scraper.go
```diff
@@ -90,44 +95,144 @@ func (s *dcgmScraper) scrape(_ context.Context) (pmetric.Metrics, error) {
...
+		if metric, ok := metrics["DCGM_FI_DEV_ENC_UTIL"]; ok {
+			encUtil := float64(metric.asInt64()) / 100.0 /* normalize */
+			s.mb.RecordGpuDcgmCodecEncoderUtilizationDataPoint(now, encUtil)
+		}
+		if metric, ok := metrics["DCGM_FI_DEV_DEC_UTIL"]; ok {
+			decUtil := float64(metric.asInt64()) / 100.0 /* normalize */
+			s.mb.RecordGpuDcgmCodecDecoderUtilizationDataPoint(now, decUtil)
+		}
+		if metric, ok := metrics["DCGM_FI_DEV_FB_FREE"]; ok {
+			bytesFree := 1e6 * metric.asInt64() /* MBy to By */
```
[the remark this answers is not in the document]
igorpeshansky: None of this is very well documented, but yes, this has the type `DCGM_FT_INT64`, and is thus measured in whole megabytes. The type was determined empirically.
Sadly, we can't use `asFloat64` here, because that would just give us nonsense[^1].

[^1]: https://github.com/NVIDIA/go-dcgm/blob/main/pkg/dcgm/fields.go#L189-L191

receiver/dcgmreceiver/scraper.go
```diff
@@ -90,44 +95,144 @@ func (s *dcgmScraper) scrape(_ context.Context) (pmetric.Metrics, error) {
...
+		if metric, ok := metrics["DCGM_FI_PROF_PCIE_TX_BYTES"]; ok {
+			pcieTx := int64(float64(metric.asInt64()) * (s.config.CollectionInterval.Seconds())) /* rate to delta */
+			s.mb.RecordGpuDcgmPcieIoDataPoint(now, pcieTx, metadata.AttributeNetworkIoDirectionTransmit)
+		}
+		if metric, ok := metrics["DCGM_FI_PROF_PCIE_RX_BYTES"]; ok {
+			pcieRx := int64(float64(metric.asInt64()) * (s.config.CollectionInterval.Seconds())) /* rate to delta */
+			s.mb.RecordGpuDcgmPcieIoDataPoint(now, pcieRx, metadata.AttributeNetworkIoDirectionReceive)
+		}
+		if metric, ok := metrics["DCGM_FI_PROF_NVLINK_TX_BYTES"]; ok {
+			nvlinkTx := int64(float64(metric.asInt64()) * (s.config.CollectionInterval.Seconds())) /* rate to delta */
```
quentinmit, line 174: This only works if `scrape()` is called exactly once per collection interval. 

It sounds like the API already gives you timestamps for each point, so you can walk the list of points and integrate each point with the corresponding timestamp. And you probably need to do that from a separate collection goroutine if it needs to happen more often than the collection interval.

receiver/dcgmreceiver/internal/metadata/generated_metrics.go
```diff
@@ -653,44 +1177,84 @@ func (mb *MetricsBuilder) Emit(rmo ...ResourceMetricsOption) pmetric.Metrics {
...
+}
+
+// RecordGpuDcgmMemoryBytesUsedDataPoint adds a data point to gpu.dcgm.memory.bytes_used metric.
+func (mb *MetricsBuilder) RecordGpuDcgmMemoryBytesUsedDataPoint(ts pcommon.Timestamp, val int64, gpuMemoryStateAttributeValue AttributeGpuMemoryState) {
+	mb.metricGpuDcgmMemoryBytesUsed.recordDataPoint(mb.startTime, ts, val, gpuMemoryStateAttributeValue.String())
+}
+
+// RecordGpuDcgmNvlinkIoDataPoint adds a data point to gpu.dcgm.nvlink.io metric.
+func (mb *MetricsBuilder) RecordGpuDcgmNvlinkIoDataPoint(ts pcommon.Timestamp, val int64, networkIoDirectionAttributeValue AttributeNetworkIoDirection) {
+	mb.metricGpuDcgmNvlinkIo.recordDataPoint(mb.startTime, ts, val, networkIoDirectionAttributeValue.String())
```
quentinmit, line 1222: This is using the wrong start time if these are deltas. What happens when these incorrect delta points get written to Cloud Monitoring?

receiver/dcgmreceiver/util.go
```diff
@@ -39,44 +48,44 @@ func (m *dcgmMetric) asInt64() int64 {
 	return *(*int64)(unsafe.Pointer(&m.value[0]))
```
quentinmit, line 48: As the name implies, this is unsafe. Accessing a pointer in this way is undefined behavior in Go.

You want `int64(binary.NativeEndian.Uint64(m.value[:8]))`, or for processing a whole slice at a time, `binary.Read(...)`. Ditto for the three functions above.

receiver/dcgmreceiver/scraper.go
```diff
@@ -90,44 +95,144 @@ func (s *dcgmScraper) scrape(_ context.Context) (pmetric.Metrics, error) {
...
+		if metric, ok := metrics["DCGM_FI_DEV_FB_FREE"]; ok {
+			bytesFree := 1e6 * metric.asInt64() /* MBy to By */
+			s.mb.RecordGpuDcgmMemoryBytesUsedDataPoint(now, bytesFree, metadata.AttributeGpuMemoryStateFree)
+		}
+		if metric, ok := metrics["DCGM_FI_DEV_FB_USED"]; ok {
+			bytesUsed := 1e6 * metric.asInt64() /* MBy to By */
+			s.mb.RecordGpuDcgmMemoryBytesUsedDataPoint(now, bytesUsed, metadata.AttributeGpuMemoryStateUsed)
+		}
+		if metric, ok := metrics["DCGM_FI_DEV_FB_RESERVED"]; ok {
+			bytesFree := 1e6 * metric.asInt64() /* MBy to By */
```
LujieDuan, line 156: Nit: naming - `bytesReserved`

receiver/dcgmreceiver/scraper.go
igorpeshansky: Right now, `scrape()` _is_ called exactly once. I agree that the accuracy of the current approach is somewhat suboptimal, but I did fix a few bugs, and wasn't able to get the separate collection goroutine working. We can revisit this on the [igorpeshansky-dcgm-new-metrics-parallel](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/compare/igorpeshansky-dcgm-new-metrics...igorpeshansky-dcgm-new-metrics-parallel) branch, but for now I'll stay with one scrape per collection.

receiver/dcgmreceiver/internal/metadata/generated_metrics.go
igorpeshansky: This seems to be a bug in `mdatagen`. I've changed all of these metrics to cumulative, for which this start time is more reasonable.

receiver/dcgmreceiver/util.go
igorpeshansky: Found an even more robust solution. There is no more usage of `unsafe` in this receiver. Can't speak for [`go-dcgm`](https://github.com/NVIDIA/go-dcgm/blob/f83cdef499b2cbaba53ab36257e492cfbac0ab60/pkg/dcgm/fields.go#L185-L191), though…

quentinmit (approved):

PR #223, opened 2024-08-07
Collect metrics asynchronously

quentinmit:


receiver/dcgmreceiver/scraper_gpu_test.go
```diff
@@ -198,24 +213,37 @@ func TestScrapeOnProfilingPaused(t *testing.T) {
...
-	err := testprofilepause.PauseProfilingMetrics()
-	if err != nil {
-		if errors.Is(err, testprofilepause.FeatureNotSupportedError) {
-			t.Skipf("Pausing profiling not supported")
-		} else {
-			t.Errorf("Pausing profiling failed with error %v", err)
+	defer func() {
+		client, err := scraper.initClient()
+		require.NoError(t, err)
+		defer client.cleanup()
```
igorpeshansky, line 219: I'm confused — why do we need this? Is it because we don't want to access `dcgm.handle` across threads? Or does that get overwritten somehow?

I actually suspect it might be enough to call `dcgmInit` (or even `dcgm.Init`) here:
```suggestion
		dcgmCleanup, err := dcgm.Init(dcgm.Standalone, config.TCPAddrConfig.Endpoint, "0")
		require.NoError(t, err)
		defer dcgmCleanup()
```

receiver/dcgmreceiver/util.go
```diff
@@ -19,39 +19,99 @@ package dcgmreceiver
...
-
-// rateIntegrator converts timestamped values that represent rates into
-// cumulative values. It assumes the rate stays constant since the last
-// timestamp.
-type rateIntegrator[V int64 | float64] struct {
-	lastTimestamp    int64
-	aggregatedRateUs V // the integration of the rate over microsecond timestamps.
+// For each metric, we need to track:
+type metricStats struct {
+	// Timestamp (us)
```
igorpeshansky, line 28: You used unicode in other comments…
```suggestion
	// Timestamp (µs)
```

receiver/dcgmreceiver/util.go
```diff
@@ -19,39 +19,99 @@ package dcgmreceiver
...
+	// Last value (for gauge metrics), as int or float
```
igorpeshansky, line 29:
```suggestion
	// Last value (for gauge metrics), as int64 or float64
```

receiver/dcgmreceiver/util.go
```diff
@@ -19,39 +19,99 @@ package dcgmreceiver
...
+	metric, ok := m[name]
+	if ok && metric.lastFieldValue != nil {
+		return asInt64(*metric.lastFieldValue)
+	}
+	return 0, false
+}
+func (m MetricsMap) IntegratedRate(name string) (int64, bool) {
+	metric, ok := m[name]
+	if ok {
+		return metric.integratedRateSeconds, true
```
igorpeshansky, line 105: Just realized — this is fine for the ops agent, as its collection interval is always in whole seconds, but if we ever submit this upstream, nothing restricts the OTel collector interval, so this could be wrong. Let's add a comment documenting this assumption?

receiver/dcgmreceiver/util.go
```diff
@@ -19,39 +19,99 @@ package dcgmreceiver
...
+}
+func (m MetricsMap) CumulativeTotal(name string) (int64, bool) {
+	metric, ok := m[name]
+	if ok {
+		return metric.cumulativeValue, true
+	}
+	return 0, false
 }
 
 type defaultMap[K comparable, V any] struct {
```
igorpeshansky, line 117: Isn't this now unused as well?

receiver/dcgmreceiver/util_test.go
```diff
@@ -18,103 +18,118 @@
...
+	testMetricStatsCumulative[int64](t, fieldValueInt64)
 }
 
-func TestCumulativeTrackerFloat64(t *testing.T) {
-	testCumulativeTracker[float64](t)
+func TestMetricStatsCumulativeFloat64(t *testing.T) {
+	testMetricStatsCumulative[float64](t, fieldValueFloat64)
 }
 
 func TestDefaultMap(t *testing.T) {
```
igorpeshansky, line 135: Also unused — let's remove?

receiver/dcgmreceiver/client.go
```diff
@@ -372,83 +361,79 @@ func (client *dcgmClient) cleanup() {
...
+		if fieldValue.EntityGroupId != dcgm.FE_GPU {
+			continue
+		}
+		gpuIndex := fieldValue.EntityId
+		if _, ok := client.devices[gpuIndex]; !ok {
+			device, err := newDeviceMetrics(client.logger, gpuIndex)
+			if err != nil {
+				continue
+			}
+			client.devices[gpuIndex] = device
```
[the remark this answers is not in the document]
quentinmit: What I have (early exit) is preferred by Go style: https://go.dev/doc/effective_go#if

receiver/dcgmreceiver/client.go
```diff
@@ -139,30 +138,20 @@ func initializeDcgm(endpoint string, logger *zap.Logger) (func(), error) {
 	return dcgmCleanup, nil
 }
 
-func discoverDevices(logger *zap.Logger) ([]uint, []string, []string, error) {
-	supportedDeviceIndices, err := dcgm.GetSupportedDevices()
+func newDeviceMetrics(logger *zap.SugaredLogger, gpuIndex uint) (deviceMetrics, error) {
```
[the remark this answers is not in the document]
quentinmit: `new` in a function name doesn't imply a pointer in Go; see https://go.dev/doc/effective_go#package-names

receiver/dcgmreceiver/client.go
```diff
@@ -139,30 +138,20 @@ func initializeDcgm(endpoint string, logger *zap.Logger) (func(), error) {
...
-		UUIDs = append(UUIDs, deviceInfo.UUID)
-		logger.Sugar().Infof("Discovered NVIDIA device %s with UUID %s", names[gpuIndex], UUIDs[gpuIndex])
+	device := deviceMetrics{
+		ModelName: deviceInfo.Identifiers.Model,
+		UUID:      deviceInfo.UUID,
+		Metrics:   MetricsMap{},
 	}
-
-	return devices, names, UUIDs, nil
+	logger.Infof("Discovered NVIDIA device %s with UUID %s", device.ModelName, device.UUID)
```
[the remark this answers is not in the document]
quentinmit: Done.

receiver/dcgmreceiver/client.go
```diff
@@ -372,83 +361,79 @@ func (client *dcgmClient) cleanup() {
...
+		}
+		gpuIndex := fieldValue.EntityId
+		if _, ok := client.devices[gpuIndex]; !ok {
+			device, err := newDeviceMetrics(client.logger, gpuIndex)
+			if err != nil {
+				continue
+			}
+			client.devices[gpuIndex] = device
+		}
+		device := client.devices[gpuIndex]
```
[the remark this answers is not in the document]
quentinmit: Yeah, you're not even copying the model and UUID strings, just the pointers to them (so this copy is ~6 pointers)

receiver/dcgmreceiver/client_gpu_test.go
```diff
@@ -135,41 +138,55 @@ func TestNewDcgmClientWithGpuPresent(t *testing.T) {
...
+		time.Sleep(client.pollingInterval)
+	}
+	deviceMetrics := client.devices
+	expectedMetrics := LoadExpectedMetrics(t, client.devices[0].ModelName)
 
-	asFloat64 := func(metric dcgmMetric) float64 {
-		require.IsTypef(t, float64(0), metric.value, "Unexpected metric type: %T", metric.value)
-		value, _ := metric.value.(float64)
+	lastFloat64 := func(metric *metricStats) float64 {
+		assert.Equal(t, dcgm.DCGM_FT_DOUBLE, metric.lastFieldValue.FieldType, "Unexpected metric type: %+v", metric.lastFieldValue)
```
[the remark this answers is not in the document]
quentinmit: I think `assert.Equal` is fine; the test will be marked as failing, `asFloat64` will then do a cast and we'll get to check the value still even if the type is unexpected. If it's not castable to `float64` we'll fail at the `require.True` below.

receiver/dcgmreceiver/scraper.go
```diff
@@ -220,188 +177,180 @@ func discoverRequestedFields(config *Config) []string {
...
+func (s *dcgmScraper) run(ctx context.Context, metricsCh chan<- map[uint]deviceMetrics) error {
+	defer close(metricsCh)
+	for {
+		client, _ := s.initClient()
+		// Ignore the error; it's logged in initClient.
+		if client != nil {
+			s.runOnce(ctx, client, metricsCh)
+		}
+		select {
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
quentinmit: `ctx.Done()` returns a channel that is closed when the context is done. A receive on a closed channel is defined to always return a zero value, no matter how many times it is called: https://go.dev/ref/spec#Receive_operator

receiver/dcgmreceiver/scraper.go
```diff
@@ -220,188 +177,180 @@ func discoverRequestedFields(config *Config) []string {
...
+			return ctx.Err()
+		case metricsCh <- map[uint]deviceMetrics{}:
```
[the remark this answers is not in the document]
quentinmit: The first usable send/receive is executed, starting at the top of the `select` statement. So if the context is done and `metricsCh` is sendable, it will return and never send on `metricsCh`. (Then the `defer close(metricsCh)` will unblock any receivers, as discussed above.)

receiver/dcgmreceiver/scraper_gpu_test.go
quentinmit: This is needed because there is no active client at the time we're trying to pause/resume profiling.

You're right that a bare dcgm.Init/cleanup should be sufficient, but I was trying to avoid duplicating the knowledge of how to init here in the test. WDYT?

receiver/dcgmreceiver/client.go
```diff
@@ -359,7 +348,7 @@ func setWatchesOnEnabledFields(pollingInterval time.Duration, logger *zap.Logger
 		// Note: DCGM retained samples = Max(maxKeepSamples, maxKeepTime/updateFreq)
 		updateFreqUs:   int64(pollingInterval / time.Microsecond),
 		maxKeepTime:    600.0, /* 10 min */
-		maxKeepSamples: int32(15),
+		maxKeepSamples: int32(100),
```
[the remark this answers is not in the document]
quentinmit: Good question. TODO added but I'll keep the comment thread open until we address it.

receiver/dcgmreceiver/client.go
```diff
@@ -372,83 +361,79 @@ func (client *dcgmClient) cleanup() {
...
-				msg := fmt.Sprintf("Unable to poll DCGM daemon for GPU %d on %s", gpuIndex, pollErr)
-				client.issueWarningForFailedQueryUptoThreshold(gpuIndex, "all-profiling-metrics", msg)
-				err.AddPartial(1, fmt.Errorf("%s", msg))
-			}
-		}
+// collect will poll dcgm for any new metrics, updating client.devices as appropriate
+// It returns the estimated polling interval.
+func (client *dcgmClient) collect() (time.Duration, error) {
+	client.logger.Debugf("Polling DCGM daemon for field values")
+	fieldValues, pollTime, err := dcgmGetValuesSince(dcgm.GroupAllGPUs(), client.enabledFieldGroup, client.lastSuccessfulPoll)
```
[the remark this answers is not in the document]
quentinmit: My thought was that if dcgm gives us data for a new GPU we'll handle it correctly, so it's harmless whether we get data for other GPUs or not.

receiver/dcgmreceiver/scraper.go
```diff
@@ -220,188 +177,180 @@ func discoverRequestedFields(config *Config) []string {
 	return requestedFields
 }
 
-func (s *dcgmScraper) scrape(_ context.Context) (pmetric.Metrics, error) {
-	err := s.initClient()
-	if err != nil || s.client == nil {
-		return s.mb.Emit(), err
+func (s *dcgmScraper) run(ctx context.Context, metricsCh chan<- map[uint]deviceMetrics) error {
```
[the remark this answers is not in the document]
quentinmit: `run` is common for Goroutine entrypoint methods, but I can't find any explicit guidance on naming. I made these slightly longer.

receiver/dcgmreceiver/scraper.go
```diff
@@ -220,188 +177,180 @@ func discoverRequestedFields(config *Config) []string {
...
 
-	s.settings.Logger.Sugar().Debug("Client created, collecting metrics")
-	deviceMetrics, err := s.client.collectDeviceMetrics()
-	if err != nil {
-		s.settings.Logger.Sugar().Warnf("Metrics not collected; err=%v", err)
-		return s.mb.Emit(), err
+func (s *dcgmScraper) runOnce(ctx context.Context, client *dcgmClient, metricsCh chan<- map[uint]deviceMetrics) {
+	defer client.cleanup()
+	for {
+		waitTime, err := client.collect()
```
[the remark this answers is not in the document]
quentinmit: `collect()` returns the duration of data it scraped, as a hint for the next poll interval. Then we do the subsequent math on it, to produce the time we're going to wait. I reused the same variable but I suppose I could split that into two names.

I think this math is actually broken now, BTW, because we now pass `client.lastSuccessfulPoll` to `dcgm.GetValuesSince`. So if the first poll returns 10s of data, the math will compute a 5s wait, it will then wait 5s and poll again, at which point we'll get 5s of data and the math will compute a 2.5s wait, etc. I think we probably need to compute the expected wait time from `maxKeepSamples`/`updateFreqUs` instead.

receiver/dcgmreceiver/util.go
quentinmit: Hm, I don't think this is wrong? It quantizes to the nearest unit-second, but since the original unit is per-second, that means it will just quantize to the nearest unit (e.g. bytes/second-seconds -> nearest byte). That seems correct to me regardless of whether the collection interval is in whole seconds, no? At worst it's just going to result in aliasing, but that's unavoidable if the cumulative is an INT64.

receiver/dcgmreceiver/util_test.go
quentinmit: How is this unused? This is testing to make sure that we correctly calculate the cumulative value of a float64 metric.  It seems fine to me to keep that test, whether or not we're currently scraping a metric of that shape.

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
igorpeshansky: I'm more worried about errors when trying to query data from unsupported GPUs (I mean, they must have been excluded from the list of supported devices for a reason, right?).

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
igorpeshansky: Heh. Maybe I'm missing something, but `deviceMetrics.ModelName` and `deviceMetrics.UUID` are `string`s, not `*string`s… It shouldn't be a big deal to copy them, regardless.

receiver/dcgmreceiver/scraper.go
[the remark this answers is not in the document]
igorpeshansky: Maybe we just need to apply `/ 2` to `s.config.CollectionInterval`? I.e.:
```go
		waitTime = max(
			100*time.Millisecond,
			min(s.config.CollectionInterval/2, waitTime),
		)
```
?

receiver/dcgmreceiver/util.go
igorpeshansky: Right, yes. It was confusing, so maybe we need a comment explaining this in more detail? Up to you.

receiver/dcgmreceiver/scraper_gpu_test.go
igorpeshansky: (a) Then you shouldn't need that for resuming, only for pausing.
(b) Should we instead wait for the client to be created before trying to pause profiling?

receiver/dcgmreceiver/util_test.go
igorpeshansky: I meant that the `defaultMap` functionality is unused, and thus we don't need `TestDefaultMap`.

receiver/dcgmreceiver/scraper.go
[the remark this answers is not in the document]
quentinmit: No, because that will still inch smaller over time (10s poll returns 9.9s of data, 9.9s poll returns 9.8s of data, etc.)

Also the very first poll always returns 0 data, so the first `waitTime` is going to be 100ms.

receiver/dcgmreceiver/scraper_gpu_test.go
quentinmit: a) The resume happens in a defer, which means after `scraper.stop` had been called and the client has been uninitialized
b) No, because this is trying to test that fewer metrics are collected... if we delay pausing until after the client is created, it will see those metrics from before the pause.

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
quentinmit: We already have to worry about that if there are two GPUs with different sets of supported metrics, so assuming that works, I think it should also be fine if the new GPU has fewer supported metrics.

receiver/dcgmreceiver/util.go
quentinmit: Comment added.

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
quentinmit: This results in a 10s buffer, so we poll every 5s. IMO that seems reasonable?

receiver/dcgmreceiver/scraper.go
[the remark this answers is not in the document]
quentinmit: I've updated the code to return `updateFreqUs * maxKeepSamples` as the buffer size, which with the current knobs means we'll collect every 5s.

receiver/dcgmreceiver/scraper_gpu_test.go
```diff
@@ -61,6 +61,9 @@ func TestScrapeWithGpuPresent(t *testing.T) {
 	err := scraper.start(context.Background(), componenttest.NewNopHost())
 	require.NoError(t, err)
 
+	// Make sure we wait long enough to do multiple collections.
+	time.Sleep(10 * time.Second)
```
igorpeshansky, line 65: Should this be a constant somewhere?

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
igorpeshansky: The polling interval is 1s, isn't it? Would that make it 100s buffer? Or did I miss that each update here collects multiple samples?

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
igorpeshansky: That's not what I was saying. Suppose one of the devices is completely unsupported. Would attempting to query that device trigger a library error, rather than a sequence of "unsupported" values?

receiver/dcgmreceiver/scraper_gpu_test.go
igorpeshansky: (a) Maybe this defer should happen in a smaller scope, before `scraper.stop()` is called?
(b) The test is probably just fragile. Doesn't the client collect new values since the last poll? Should we change it to return no points if no new values appeared since last scrape? Then this test will do the right thing (create scraper, trigger a scrape, pause profiling, trigger another scrape, check metrics from the last scrape, resume profiling, stop scraping).

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
quentinmit: scraper.go has:
```go
const scrapePollingInterval = 100 * time.Millisecond // TODO: Choose an appropriate value
```
So the total buffer dcgm will retain is `100 * 100ms`, or `10s`.

receiver/dcgmreceiver/scraper_gpu_test.go
quentinmit: This test has always paused profiling, then started scraping, then checked metrics, then stopped scraping, then resumed profiling. 

Are you suggesting that the test should be more complicated? We talked briefly about that, but it's not clear to me what behavior we should actually be testing for in that case.

receiver/dcgmreceiver/scraper_gpu_test.go
quentinmit: It's only used in the test, so it's fine for it to be a constant. But, there's a broader point that the test should be more explicit about what it's checking. I refactored this into two separate tests, one of which explicitly checks to make sure we're not collecting too often.

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
quentinmit: That's certainly my assumption. At the top level API, there's separate `dcgmGetAllDevices` and `dcgmGetAllSupportedDevices` functions. But there's only one enum, `DCGM_GROUP_ALL_GPUS`. The docs for `dcgmGetValuesSince_v2` just say

> Alternatively, pass in the group id as `DCGM_GROUP_ALL_GPUS` to perform operation on all the GPUs or `DCGM_GROUP_ALL_NVSWITCHES` to perform the operation on all NvSwitches.

If that fails if there's also an unsupported GPU in the system, that seems like it would be unusable. Also, `DCGM_ST_NOT_SUPPORTED` is not documented as a possible return value for `dcgmGetValuesSince_v2` (which it is for other functions), so I think that also indicates it's okay to call with unsupported GPUs present.

If you want a reliable answer, we'll have to ask NVIDIA.

receiver/dcgmreceiver/scraper.go
```diff
@@ -216,14 +216,19 @@ func (s *dcgmScraper) pollClient(ctx context.Context, client *dcgmClient, metric
...
-		case <-time.After(waitTime):
+		after := time.After(waitTime)
+		for after != nil {
+			deviceMetrics := client.getDeviceMetrics()
+			select {
+			case <-ctx.Done():
+				return
+			case <-collectTriggerCh:
+				// Loop and trigger a collect() again.
+				after = nil
```
igorpeshansky, line 227: Just for my own understanding, this is equivalent to:
```go
L:
		for {
			deviceMetrics := client.getDeviceMetrics()
			select {
			case <-ctx.Done():
				return
			case <-collectTriggerCh:
				// Loop and trigger a collect() again.
				break L
			case metricsCh <- deviceMetrics:
			case <-time.After(waitTime):
				break L
			}
		}
```
right?

receiver/dcgmreceiver/scraper.go
```diff
@@ -216,14 +216,19 @@ func (s *dcgmScraper) pollClient(ctx context.Context, client *dcgmClient, metric
...
-		select {
-		case <-ctx.Done():
-			return
-		case <-collectTriggerCh:
-			// Loop and trigger a collect() again.
-		case metricsCh <- deviceMetrics:
-		case <-time.After(waitTime):
+		after := time.After(waitTime)
+		for after != nil {
+			deviceMetrics := client.getDeviceMetrics()
```
igorpeshansky, line 221: I may have asked this before, but are we going to call `client.getDeviceMetrics()` unconditionally, even if nobody is reading from `metricsCh`? If so, can we only call it on-demand?

receiver/dcgmreceiver/scraper_gpu_test.go
```diff
@@ -48,20 +61,60 @@ func TestScrapeWithGpuPresent(t *testing.T) {
...
+	metrics, err := collectScraperResult(t, context.Background(), scraper)
 	assert.NoError(t, err)
 
-	require.NotNil(t, scraper.client)
-	require.NotEmpty(t, scraper.client.devicesModelName)
-	expectedMetrics := loadExpectedScraperMetrics(t, scraper.client.getDeviceModelName(0))
-	validateScraperResult(t, metrics, expectedMetrics)
+	assert.NoError(t, scraper.stop(context.Background()))
+
+	// We should have seen 1 initial scrape + 3 timed scrapes + 2 scrapes triggered by `collectScraperResult`.
```
igorpeshansky, line 102:
```suggestion
	// We should have seen 1 initial scrape + 3 timed scrapes + up to 2 scrapes triggered by `collectScraperResult`.
```
Only cumulative metrics trigger 2 scrapes, right?

receiver/dcgmreceiver/scraper.go
quentinmit: Almost, but your code constructs a new `time.After` channel each time it loops (potentially sleeping for more than the desired `waitTime`), whereas my code reuses the same one and will thus wait for only one instance of `waitTime`. I used my version just because loop labels look too much like goto labels and I have an aversion to them :)

receiver/dcgmreceiver/scraper.go
quentinmit: Yes, we discussed this on one of the video chats. We said that it would be premature optimization to avoid it.

There's no way in Go to "check if a channel has a waiting receiver" without trying to send on it, and to send on it you need to already have the value you're sending. To avoid that construction we'd need to introduce a higher level of synchronization (for instance, use a channel of channels).

receiver/dcgmreceiver/scraper_gpu_test.go
quentinmit: `collectScraperResult` is a test helper at the top of this file. It always scrapes until it gets a metric value (which in this particular test should always be present), and then scrapes one more time (to make sure we have values for cumulative metrics). Those are the 2 scrapes I'm counting here. `collectScraperResult` doesn't have any way of knowing if there's a missing cumulative metric or not, so it doesn't implement "up to", it just always scrapes twice.

receiver/dcgmreceiver/scraper_gpu_test.go
quentinmit: I think I understand now that you don't like how complicated the test has become. I just moved the client initialization into `testprofilepause` - PTAL?

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
quentinmit: That device group is only used in the constructor and isn't available in the rest of the client. I guess you're asking me to store it in the client object for use later.

I was intending to make this the first step towards removing the device enumeration in the constructor entirely, but it's not worth adding even more blockers to the review.

receiver/dcgmreceiver/client.go
```diff
@@ -367,7 +371,11 @@ func (client *dcgmClient) cleanup() {
 // It returns the estimated polling interval.
 func (client *dcgmClient) collect() (time.Duration, error) {
 	client.logger.Debugf("Polling DCGM daemon for field values")
-	fieldValues, pollTime, err := dcgmGetValuesSince(dcgm.GroupAllGPUs(), client.enabledFieldGroup, client.lastSuccessfulPoll)
+	if len(client.enabledFieldIDs) == 0 {
+		// Make sure we don't try to scrape without a device group (since we don't construct one when there are no enabled fields).
```
igorpeshansky, line 375: I wonder, should we even create a client when there are no enabled fields?

receiver/dcgmreceiver/scraper_gpu_test.go
igorpeshansky: This looks a lot cleaner and easier to read, thanks!

receiver/dcgmreceiver/client.go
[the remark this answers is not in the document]
igorpeshansky: Yeah, I'd rather proceed slowly on this one and confirm with NVIDIA first. Thanks!

igorpeshansky (approved):
LGTM :shipit:

receiver/dcgmreceiver/client.go
quentinmit: Great question. I'll defer that to future cleanup, though, since I'm not changing it here.

receiver/dcgmreceiver/client.go
```diff
@@ -352,103 +344,108 @@ func setWatchesOnFields(logger *zap.Logger, deviceGroup dcgm.GroupHandle, fieldI
...
+		}
+		device := client.devices[gpuIndex]
 		dcgmName := dcgmIDToName[dcgm.Short(fieldValue.FieldId)]
 		if err := isValidValue(fieldValue); err != nil {
 			msg := fmt.Sprintf("Received invalid value (ts %d gpu %d) %s: %v", fieldValue.Ts, gpuIndex, dcgmName, err)
-			client.issueWarningForFailedQueryUptoThreshold(gpuIndex, dcgmName, msg)
-			if client.retryBlankValues && errors.Is(err, errBlankValue) {
-				retry = true
-			}
+			client.issueWarningForFailedQueryUptoThreshold(fmt.Sprintf("device%d.%s", gpuIndex, dcgmName), msg)
```
igorpeshansky, line 404: On H100, we continue getting a lot of these blank value warnings, as late as 5 minutes after startup, until the threshold gets saturated. We are either querying DCGM for these fields too frequently, or there's another reason why blank values are returned. This produces a lot of log spam from the binary — I suggest we suppress blank value errors, at least for the initial ramp-up phase.

receiver/dcgmreceiver/client.go
```diff
@@ -216,17 +212,13 @@ func getSupportedProfilingFields() ([]dcgm.Short, error) {
...
 	for _, ef := range requestedFields {
-		support := false
-		for _, sf := range supportedRegularFields {
-			if sf == ef {
-				support = true
-				break
-			}
-		}
+		// For fields like `DCGM_FI_DEV_*`, which are not
+		// profiling fields, assume they are always present.
```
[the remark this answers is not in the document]
quentinmit: I don't think this is a broken assumption in context. We know there are some `DCGM_FI_DEV_*` metrics which return `field not supported` on scrape, but the way we were previously determining they were not supported is by attempting a scrape. Therefore, it is always safe to include these metrics in the scrape.

receiver/dcgmreceiver/client_gpu_test.go
```diff
@@ -285,14 +288,37 @@ func TestCollectGpuProfilingMetrics(t *testing.T) {
...
-			seenMetric[fmt.Sprintf("gpu{%d}.metric{%s}", gpuIndex, name)] = true
+			seenMetric[name]++
 		}
 	}
 
-	for gpuIndex := range deviceMetrics {
-		for _, metric := range expectedMetrics {
-			assert.True(t, seenMetric[fmt.Sprintf("gpu{%d}.metric{%s}", gpuIndex, metric)], fmt.Sprintf("%s on gpu %d", metric, gpuIndex))
+	for name, count := range seenMetric {
+		assert.Equalf(t, count, len(deviceMetrics), "metric %q found on an unexpected number of GPUs", name)
```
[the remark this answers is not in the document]
quentinmit: I don't think it's possible to get heterogeneous GPUs on GCE. If and when we start running these tests on heterogeneous GPUs, we can worry about making the test more robust then.

receiver/dcgmreceiver/client_gpu_test.go
```diff
@@ -59,51 +61,55 @@ func defaultClientSettings() *dcgmClientSettings {
...
+		}
 	}
 	for _, f := range unavailableFields {
-		unavailableFieldsString = append(unavailableFieldsString, dcgmIDToName[f])
+		name := dcgmIDToName[f]
+		if !strings.HasPrefix(name, "DCGM_FI_DEV_") {
+			unavailableFieldsString = append(unavailableFieldsString, name)
+		}
 	}
+	sort.Strings(enabledFieldsString)
```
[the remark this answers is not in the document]
quentinmit: I don't want to encode that in the test because it's not the point of the test. Checking the order would just make a more fragile test.

receiver/dcgmreceiver/client_gpu_test.go
```diff
@@ -59,51 +61,55 @@ func defaultClientSettings() *dcgmClientSettings {
...
 	require.Nil(t, err)
-	enabledFields, unavailableFields := filterSupportedFields(allFields, supportedRegularFields, supportedProfilingFields)
+	enabledFields, unavailableFields := filterSupportedFields(allFields, supportedProfilingFields)
 
 	var enabledFieldsString []string
 	var unavailableFieldsString []string
 	for _, f := range enabledFields {
-		enabledFieldsString = append(enabledFieldsString, dcgmIDToName[f])
+		name := dcgmIDToName[f]
+		if !strings.HasPrefix(name, "DCGM_FI_DEV_") {
```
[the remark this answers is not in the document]
quentinmit: Eh, I can't do that down below when I'm filtering `want`, so I might as well use the same test up here.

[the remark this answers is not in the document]
igorpeshansky: You could have checked for having the `DCGM_FI_PROF_` prefix everywhere instead, but whatever, it's a test…

receiver/dcgmreceiver/client.go
```diff
@@ -367,13 +369,13 @@ func (client *dcgmClient) getDeviceMetrics() map[uint]deviceMetrics {
...
-func (client *dcgmClient) issueWarningForFailedQueryUptoThreshold(dcgmName string, reason string) {
+func (client *dcgmClient) issueWarningForFailedQueryUptoThreshold(dcgmName string, limit int, reason string) {
 	client.deviceMetricToFailedQueryCount[dcgmName]++
 
 	failedCount := client.deviceMetricToFailedQueryCount[dcgmName]
-	if failedCount <= maxWarningsForFailedDeviceMetricQuery {
-		client.logger.Warnf("Unable to query '%s' on '%s'", dcgmName, reason)
-		if failedCount == maxWarningsForFailedDeviceMetricQuery {
+	if failedCount <= limit {
+		client.logger.Warnf("%s", reason)
```
igorpeshansky, line 377: `logger.Warnf("%s", reason)` is the same as `logger.Warn(reason)` (https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Warn). But I can fix that after we merge this PR…

receiver/dcgmreceiver/client.go
```diff
@@ -328,9 +328,11 @@ func (client *dcgmClient) collect() (time.Duration, error) {
 		if err := isValidValue(fieldValue); err == errBlankValue {
 			// Blank values are expected at startup.
 			continue
+		} else if err == errNotSupported {
+			client.issueWarningForFailedQueryUptoThreshold(dcgmName, 1, fmt.Sprintf("Field '%s' is not supported. Metric '%s' will not be collected", dcgmName, dcgmName))
```
igorpeshansky, line 332: The `Metric '%s' will not be collected` part is redundant (in the profiling metrics as well). But I can fix that after we merge this PR…

receiver/dcgmreceiver/client.go
```diff
@@ -113,7 +113,7 @@ func newClient(settings *dcgmClientSettings, logger *zap.Logger) (*dcgmClient, e
 		deviceGroup:                    deviceGroup,
 		devices:                        map[uint]deviceMetrics{},
 		lastSuccessfulPoll:             time.Now(),
-		deviceMetricToFailedQueryCount: make(map[string]uint64),
+		deviceMetricToFailedQueryCount: make(map[string]int),
```
igorpeshansky, line 116: Just curious, was this so you wouldn't have to declare `limit` as `uint64` as well?

igorpeshansky (approved):

receiver/dcgmreceiver/client.go
quentinmit: That's why I noticed, yeah, but I don't see any reason this needed to be `uint64` - in general in Go, anything that's counting should just use `int` (basically the equivalent of `size_t` in C).

PR #224, opened 2024-08-07
Update otel collector to v0.106.1.

franciscovalentecastro:
This PR updates the opentelemetry-collector dependency to `v0.106.1`. Some changes were required after the update. Some details :

* `CreateSettings` and `NewCreateNopSettings` were deprecate in favour of `Settings` and `NewNopSettings` : 
  * https://github.com/open-telemetry/opentelemetry-collector/pull/10336
* Import `golang.org/x/tools/cmd/goimports` instead of only `golang.org/x/tools` which wasn't providing the module correctly.
* `UnmarshalConfig` is deprecated. Using `Unmarshal` instead : 
  * https://github.com/open-telemetry/opentelemetry-collector/issues/7102
* No default `ConfigProvider`s are set anymore. Supplied manually `file`, `env` and `yaml` config providers which seem are the only necessary for the `opentelemetry-operations-collector`: 
  * https://github.com/open-telemetry/opentelemetry-collector/issues/10290


service/main.go
```diff
@@ -43,6 +48,15 @@ func MainContext(ctx context.Context) {
 	params := otelcol.CollectorSettings{
 		Factories: components,
 		BuildInfo: info,
+		ConfigProviderSettings: otelcol.ConfigProviderSettings{
+			ResolverSettings: confmap.ResolverSettings{
+				ProviderFactories: []confmap.ProviderFactory{
+					fileprovider.NewFactory(),
+					envprovider.NewFactory(),
+					yamlprovider.NewFactory(),
```
avilevy18, line 56: According to https://github.com/open-telemetry/opentelemetry-collector/issues/10290#issuecomment-2145370916 the default should be `"env"` (or in this case `envprovider`). 

Spoke offline with @franciscovalentecastro, and integration tests fail without the file provider.

Could there be other providers that we are might be missing that are tests are not picking up?

franciscovalentecastro: As discussed offline, a guideline to set providers here could be the `default` that was set before the update. I found `newDefaultConfigProviderSettings` was modified here https://github.com/open-telemetry/opentelemetry-collector/pull/10436. I set the defaults from there :

```
func newDefaultConfigProviderSettings(uris []string) ConfigProviderSettings {
	return ConfigProviderSettings{
		ResolverSettings: confmap.ResolverSettings{
			URIs: uris,
			ProviderFactories: []confmap.ProviderFactory{
				fileprovider.NewFactory(),
				envprovider.NewFactory(),
				yamlprovider.NewFactory(),
				httpprovider.NewFactory(),
				httpsprovider.NewFactory(),
			},
			ConverterFactories: []confmap.ConverterFactory{expandconverter.NewFactory()},
		},
	}
}
```

avilevy18 (approved):

braydonk (approved):

PR #234, opened 2024-10-22
feat: Update OTEL dependencies to v0.112.0

StefanKurek:
This changes the OTEL dependencies to v0.112.0


processor/agentmetricsprocessor/factory_test.go
```diff
@@ -21,15 +21,16 @@ import (
...
+	"go.opentelemetry.io/collector/processor/processortest"
 )
 
 func TestCreateDefaultConfig(t *testing.T) {
 	assert.NoError(t, componenttest.CheckConfigStruct(NewFactory().CreateDefaultConfig()))
 }
 
 func TestCreateProcessor(t *testing.T) {
-	mp, err := createMetricsProcessor(context.Background(), processor.CreateSettings{}, createDefaultConfig(), consumertest.NewNop())
+	cfg := createDefaultConfig().(*Config)
```
braydonk, line 32: Could you explain why this was done? I'm having trouble understanding the reason.

braydonk (commented):
Most of the changes look fine to me, just want to understand one change.

processor/agentmetricsprocessor/factory_test.go
StefanKurek: @braydonk Honestly I can't remember why I did this for the config...but good catch! Doesn't seem to be necessary. Might have came from a copy/paste while looking at existing factory tests within the latest release.

Pushing a change to just pass in `createDefaultConfig()` in the 3 files where I did this.

service/main.go
```diff
@@ -43,6 +50,17 @@ func MainContext(ctx context.Context) {
 	params := otelcol.CollectorSettings{
 		Factories: components,
 		BuildInfo: info,
+		ConfigProviderSettings: otelcol.ConfigProviderSettings{
+			ResolverSettings: confmap.ResolverSettings{
+				ProviderFactories: []confmap.ProviderFactory{
+					fileprovider.NewFactory(),
```
ridwanmsharif, line 56: Is there a reason we're adding all these providers for our distro? Not sure why these are needed now

ridwanmsharif: Don't we only use the file provider?

jefferbrecht: There was an upstream change that forces you do explicitly list providers now, I don't have a reference at hand though

ridwanmsharif: Ack, it stems from https://github.com/open-telemetry/opentelemetry-collector/pull/9513/files

We used to default to all the default providers so this change makes sense. We should separately reduce this list to just the file provider but that shouldn't block this PR

ridwanmsharif (approved):

PR #249, opened 2025-03-14
Add oauth2authextension from upstream with no code changes

jinghan-ma:
Direct copy of the upstream code of oauth2auth extension and updated the registry to point to our local copy.

This is to make it easier to diff my changes to add STS support into the extension in a followup PR.


braydonk (approved):
Approved on the basis of working for the integration with the repo, not reviewing the extension at this time

PR #252, opened 2025-03-19
Fix kokoro build container

braydonk:
The Kokoro build container did not previously copy the local components from this repo onto it. Since we will be doing that with a couple components now, the root where we run the docker build from needs to change so we can copy all the components onto the container.

The registry entry for the local `oauth2clientauth` extension also had to change. We didn't change that module's name in our local copy, so that name needs to match. Otherwise the build fails for thinking we used an internal package from the real `oauth2clientauth` module.

### DIFF
--- a/templates/google-otel/Dockerfile.kokoro_build.go.tmpl
+++ b/templates/google-otel/Dockerfile.kokoro_build.go.tmpl
@@ -18,11 +18,12 @@ ARG BUILD_CONTAINER="alpine:3"
 FROM ${BUILD_CONTAINER} AS build
 
 RUN apk --update add make curl
-RUN mkdir -p /build
 
 ARG DISTRO_ROOT
-COPY ${DISTRO_ROOT}/Makefile /build
-COPY ${DISTRO_ROOT}/manifest.yaml /build
+COPY ${DISTRO_ROOT}/{{ .Name }} /build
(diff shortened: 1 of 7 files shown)


templates/google-otel/Dockerfile.kokoro_build.go.tmpl
```diff
@@ -18,11 +18,12 @@ ARG BUILD_CONTAINER="alpine:3"
...
 RUN apk --update add make curl
-RUN mkdir -p /build
 
 ARG DISTRO_ROOT
-COPY ${DISTRO_ROOT}/Makefile /build
-COPY ${DISTRO_ROOT}/manifest.yaml /build
+COPY ${DISTRO_ROOT}/{{ .Name }} /build
+COPY ${DISTRO_ROOT}/receiver /receiver
+COPY ${DISTRO_ROOT}/processor /processor
+COPY ${DISTRO_ROOT}/extension /extension
```
LujieDuan, line 26:
```suggestion
COPY ${DISTRO_ROOT}/exporter /exporter
COPY ${DISTRO_ROOT}/extension /extension
```

LujieDuan (commented):
LGTM with one change:

LujieDuan (approved):
LGTM!

PR #260, opened 2025-03-24
Adding the otlpreceiver to otelopscol

avilevy18:
Fixes bug: b/405981614


LujieDuan (approved):

avilevy18:
> unrelated: but why are the licences not part of the templates?

looking at the Makefile, when you regenerate otelopscol it attempts to run the `ADDLICENSE` target, but doesn't check if the dependencies are installed. My original invocation of the `make regen-otelopscol` call failed exited with a non zero exit code.

In the future we should update the target to check if the dependencies to add the license are available first. (Just run `make install-tools` first)

braydonk:
The licenses are not part of the templates because addlicense will be add
licenses to the actual generated files. I've sent the tool to some external
users to experiment with, and we don't want them to have to generate files
with Google licenses.

braydonk:
We can't easily check make install tools as a dependency because the age of
the tools directory is older than the tools, and forces make to think the
tools aren't installed. So it reinstalls everything. I hadn't written it
down, but I was hoping that all devs would run `make dev-setup` before
submitting changes so everything is set up. Couldn't find a way to automate
that and hadn't written it down anywhere.

PR #263, opened 2025-03-26
Fix license for main.yml

jinghan-ma:
Use yaml comment format


braydonk:
Running "make addlicense" should catch this kind of problem in the future

PR #271, opened 2025-04-14
internal/promqle2etest: Add test for Prometheus counter going through Otel Collector to GCM (and more)

bwplotka:
This is the initial test suite we could use as an acceptance validation and understanding of the latest nuances around various ingestions paths in OSS for GCM.

One immediate use is to test tricky StartTime processing we are fixing in b/389130459

I put it out of CI for now, but we can easily set CI for this, question is do we want to run it on every PR. We do that in prometheus-engine, but you might have a different preference.

See also chained new test cases we work on e.g. https://github.com/bwplotka/otel-ops-col/pull/1

### DIFF
--- a/internal/promqle2etest/otelgcm_test.go
+++ b/internal/promqle2etest/otelgcm_test.go
@@ -26,18 +26,19 @@ import (
 )
 
 // TestPromOtelGCM_PrometheusCounter_NoCT tests a basic counter sample behaviour
-// with a known CT limitation across 3 ingestion flows:
+// with a known CT limitation across the following ingestion flows:
 // * target --PromProto--> Prometheus (referencing, ideal, OSS behaviour).
 // * target --PromProto--> Prometheus GMP fork --GCM API--> GCM.
-// * target --PromProto--> OpenTelemetry Collector (Google Operations build) --GCM API--> GCM.
+// * target --PromProto--> OpenTelemetry Collector --GCM API--> GCM.
+// * target --PromProto--> OpenTelemetry Collector (+MSTP) --GCM API--> GCM.
 //
 // The main goal is to have a basic acceptance test on the non-trivial behaviours across multiple ingestion pipelines.
 // Currently, this test is for manual run only; to run add GCM_SECRET envvar containing GCM API read and write access (and adjust timeout).
 //
 // TODO(bwplotka): In future we could add more pipelines to test and compare e.g.
-// * target --PromProto--> Prometheus vanilla --PRW 2.0--> OpenTelemetry Collector --GCM API-->GCM.
+// * target --PromProto--> Prometheus vanilla --PRW 2.0--> OpenTelemetry Collector (+MSTP) --GCM API-->GCM.
 // * target --PromProto--> Prometheus vanilla --PRW 2.0--> GCM.
-// * target --PromProto--> OpenTelemetry Collector --OTLP--> GCM.
+// * target --PromProto--> OpenTelemetry Collector (+MSTP) --OTLP--> GCM.
 func TestPromOtelGCM_PrometheusCounter_NoCT(t *testing.T) {
 	const interval = 15 * time.Second
 
@@ -65,10 +66,9 @@ func TestPromOtelGCM_PrometheusCounter_NoCT(t *testing.T) {
 		GCMSA: GCMServiceAccountOrFail(t),
 	}
 
-	// target --PromProto?--> OpenTelemetry Collector (+ MSTP) --GCM API--> GCM.
+	// target --PromProto--> OpenTelemetry Collector (+MSTP) --GCM API--> GCM.
 	// Similar to `otel-gcm` but we use a new metricstarttimeprocessor (MSTP) processor
-	// to adjust counter samples without CT before going to GMP exporter.
-	// TODO(bwplotka): Create another Otel backend test but with MSTP and OTLP.
+	// to adjust counter samples without CT.
 	otelMSTPGCM := OtelGCMBackend{
 		Name: "otel-mstp-gcm",
 		// TODO(bwplotka): Replace with upstream once https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38594 is merged.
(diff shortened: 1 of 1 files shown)


bwplotka:
At this point I still can't get otel-gcm pipeline to work, investigating. Somehow it does not send data to GCM.

internal/promqle2etest/backend_otelgcm.go
```diff
@@ -0,0 +1,196 @@
...
+    # drop metrics if memory usage gets too high
+    check_interval: 1s
+    limit_percentage: 65
+    spike_limit_percentage: 20
+
+# Note that the googlemanagedprometheus exporter block is intentionally blank
+exporters:
+  googlemanagedprometheus:
+    project: "%s"
+    # TODO(bwplotka): Do we need to disable cumulative_normalization explicitly?
```
ridwanmsharif, line 110: We should allow the test to pass in a config imo. We want to test using the `googlemanagedprometheus` exporter but we also want to test that the `otlp` exporter works the same way when used with the `metricstarttime` processor

internal/promqle2etest/backend_otelgcm.go
```diff
@@ -0,0 +1,196 @@
...
+    metric:
+      # TODO(bwplotka): Update docs? Change defaults? Those metrics are a bad idea to include (IMO).
+      # We are working in Prometheus OSS to slowly promote more resource attr as labels (if not all one day).
```
ridwanmsharif, line 113: Thats an interesting development, have a link handy for these discussions?

internal/promqle2etest/backend_otelgcm.go
bwplotka: Yea, totally, I just want to make sure GMP path works too, something timeouts.

bwplotka: Do you have a container image that is built on top of your PR I could use?  https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38594

bwplotka:
Ok, so generally this works in ~26 secs.

I had to adjust `otel-gcm` labels to expect `instance="..."` label, somehow otel code ignores my relabelling to drop this label, perhaps some magic with `service.instance.id` is happening. 

I also added TODOs for new pipeline with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38594 or so @ridwanmsharif  - feel free to edit this PR/etc 

Closing laptop for today, let me know tomorrow where I can continue with (how to build image from your PR?)

bwplotka:
We could also merge if you prefer, this should be now good for review, even with TODOs. Test should pass for anyone with the right GCM SA for their GCP project.

internal/promqle2etest/backend_otelgcm.go
bwplotka: Some people still like joining though, so we can keep it by default, just it broke for me (int vs double descriptor issue)

bwplotka:
Added more test data for counters. 

Otel collector -> GCM without start time flow seems vary vs Prometheus and Prometheus fork completely:

image


However those are a tricky corner cases. Maybe some misconfiguration too? Added current values as "expected for `otel-gcm` to show the issue". Also we might accept this and improve it for OTLP flow @ridwanmsharif is working on. Up to us.

ridwanmsharif:
I'll give this a review but +1 to getting this merged and then iterating to figure out why these discrepancies exist.

To answer some questions now though:

> somehow otel code ignores my relabelling to drop this label, perhaps some magic with service.instance.id is happening.

The prometheus receiver sets some values for resource labels here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/prom_to_otlp.go#L47-L72
There is some more logic about when there is no `instance` label found which is probably the culprit: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d6128eb77788b005f19c31bc5d339548cfa2e64c/receiver/prometheusreceiver/internal/transaction.go#L475-L505

> Maybe some misconfiguration too?

sadly your OTel configuration looks right - you may have just unearthed a bug :)

ridwanmsharif (approved):

bwplotka:
Thanks and thanks for pre-build image @ridwanmsharif 

With only like ~10m I was able to create a test case for MSTP (metricstarttimeprocessor) + googlemanagedprometheus exporter. I don't want to pollute this base PR if we want to merge it, so I chained my next PR: https://github.com/bwplotka/otel-ops-col/pull/1

As the next steps we could add OTLP case too, but I didn't yet dive into nuances of OTLP GCM and access for it.

Results are pretty good (details in the PR), so I think we could claim googlemanagedprometheus exporter start time adjustment as "broken/deprecated" and simply recommend MSTP with your fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38594

bwplotka:
I found why currently recommended flow is "broken" -- the reason is the same as why it's broken for OTLP (the default `prometheusreceiver` broken true reset strategy adjuster -- (b/389130459)).

Luckily we address that with MSTP https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38594#discussion_r2048604963 and this might be the "short" term fix too for everyone.

I adjusted the test commentary to explain all and I merged the MSTP case too to this PR.

cc @ridwanmsharif

internal/promqle2etest/backend_otelgcm.go
```diff
@@ -0,0 +1,225 @@
...
+		return e2emon.AsInstrumented(e2e.NewFailedRunnable(name, fmt.Errorf("create otel config failed: %w", err)), "http")
+	}
+
+	if err := os.WriteFile(filepath.Join(f.Dir(), "gcm-sa.json"), gcmSA, 0600); err != nil {
+		return e2emon.AsInstrumented(e2e.NewFailedRunnable(name, fmt.Errorf("write gcm SA failed: %w", err)), "http")
+	}
+
+	args := map[string]string{
+		"--config": filepath.Join(f.Dir(), "config.yml"),
+		"--feature-gates=receiver.prometheusreceiver.EnableNativeHistograms": "",
```
dashpole, line 171: did you need to enable the `receiver.prometheusreceiver.RemoveStartTimeAdjustment` feature gate as well?

internal/promqle2etest/backend_otelgcm.go
```diff
@@ -0,0 +1,225 @@
...
+  memory_limiter:
+    # drop metrics if memory usage gets too high
+    check_interval: 1s
+    limit_percentage: 65
+    spike_limit_percentage: 20
+  %s
+
+# Note that the googlemanagedprometheus exporter block is intentionally blank
+exporters:
+  googlemanagedprometheus:
```
dashpole, line 134: We should consider testing against the OTLP endpoint.  Could be done in a follow-up, or once it is out of preview.

PR #272, opened 2025-04-14
distrogen: add hotfix for segfault

ridwanmsharif:


cmd/distrogen/distribution_test.go
```diff
@@ -53,7 +53,11 @@ func testGeneratorCase(t *testing.T, registry *Registry, testFolder string) {
 
 	g, err := NewDistributionGenerator(d, registry, true)
 	assert.NilError(t, err)
-	g.CustomTemplatesDir = os.DirFS(filepath.Join(testdataFullPath, testFolder, "templates"))
+	customTemplates := filepath.Join(testdataFullPath, testFolder, "templates")
+	_, err = os.Stat(customTemplates)
+	if err == nil {
+		g.CustomTemplatesDir = os.DirFS(customTemplates)
+	}
```
avilevy18, line 60: I think we should inverse the error checking.
Something like:
```
assert.NilError(t, err)
g.CustomTemplatesDir = os.DirFS(customTemplates)
```

avilevy18: Gotcha, LGTM them, maybe a comment would helpful?

avilevy18 (approved):

PR #275, opened 2025-04-15
samples: add samples for pre-defined metrics

ridwanmsharif:
Note that some of the configs here are invalid given the new components
we added. We are fixing these in followups.


LujieDuan (approved):
LGTM!

braydonk:
I would recommend including a README in this folder that indicates how this is only useful for internal Google teams. Many of these metrics are a large overcomplication for the general user so we want to ensure public users don't try to use these.

PR #283, opened 2025-04-22
Secret provider

XuechunHou:


specs/google-built-opentelemetry-collector.yaml
```diff
@@ -121,3 +121,4 @@ components:
     - http
     - https
     - yaml
+    - googlesecretsprovider
```
braydonk, line 124: Is this being added to `otelopscol` or `google-built-opentelemetry-collector`? I think we wanted this in `otelopscol` (the one used in Ops Agent) right?

XuechunHou: We eventually want both `otelopscol` and `google-built-opentelemetry-collector` to allow this new provider component.

PR #284, opened 2025-04-23
[otelopscol] Create custom `transformprocessor` with `ExtractPatternsRubyRegex` OTTL func linking static C library.

franciscovalentecastro:
This PR creates a custom `transformprocessor` which exactly mirrors the contrib upstream [^1] processor and adds the custom `ExtractPatternsRubyRegex` OTTL function using the `WithLogFunctions` factory options for the processor.

Some details about this PR :
- To implement `ExtractPatternsRubyRegex` the dependencies on `go-onigmo` and `onigmo` where added.
- Eventually using `oniguruma` (better maintained) + `self written go binding` would be best, but would rather do that in a followup PR.
- To build and link `onigmo` (oniguruma) in windows with CGO it seems the best approach is to use MingGW. MSVC doesn't seem to be supported very well with CGO.
- I'm still unsure how to link `onigmo` as default every time `go test` is executed for `ExtractPatternsRubyRegex`.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor


google-built-opentelemetry-collector/README.md
```diff
@@ -50,7 +50,7 @@ The Google-Built OpenTelemetry Collector is an open-source, production-ready bui
 | resource | [docs](https://www.github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourceprocessor/README.md) |
 | resourcedetection | [docs](https://www.github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/README.md) |
 | tailsampling | [docs](https://www.github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/tailsamplingprocessor/README.md) |
-| transform | [docs](https://www.github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor/README.md) |
+| transform | [docs](https://www.github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/agenttransformprocessor/README.md) |
```
quentinmit, line 53: I've gone back and forth on whether I think you should replace the `transform` processor or add a new `agenttransform` processor. Either way this change is wrong (it calls it `transform` but points at an upstream directory that doesn't exist)

otelopscol/Makefile
```diff
@@ -74,14 +77,17 @@ collector-build: clean-collector $(COLLECTOR_BINARY_NAME)
 
 # Using OCB and the local manifest file, build a collector.
 $(COLLECTOR_BINARY_NAME): $(OCB_BIN)
+	$(MAKE) install-onigmo
```
quentinmit, line 80: Using a `make` target for this means that `go build` will not be able to build the processor. I'd like us to try hard to make this work without any changes to the Makefile at all. (This probably means including the source files in our Go package so that `go build` builds them directly, or at least changing the `ldflags`/`cflags` comments in the cgo source file.)

otelopscol/Makefile
```diff
@@ -74,14 +77,17 @@ collector-build: clean-collector $(COLLECTOR_BINARY_NAME)
 
 # Using OCB and the local manifest file, build a collector.
 $(COLLECTOR_BINARY_NAME): $(OCB_BIN)
```
quentinmit, line 79:
```
$(COLLECTOR_BINARY_NAME): $(OCB_BIN) install-onigmo
```
is how you would trigger the new target as a dependency

processor/agenttransformprocessor/config.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package agenttransformprocessor
+
+import "github.com/open-telemetry/opentelemetry-collector-contrib/processor/transformprocessor"
+
+type Config struct {
```
quentinmit, line 19: I don't think you need a new type at all (just use the upstream type?)

processor/agenttransformprocessor/internal/ottlfuncs/func_extract_patterns_ruby_regex.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return func(ctx context.Context, tCtx K) (any, error) {
+		val, err := target.Get(ctx, tCtx)
+		if err != nil {
+			return nil, err
+		}
+
+		matches := r.FindStringSubmatch(val)
+		if matches == nil {
+			return pcommon.NewMap(), nil
```
quentinmit, line 61: Can/should we return nil instead of an empty map?

google-built-opentelemetry-collector/README.md
franciscovalentecastro: Removed the READMED updates. I deferred back to calling it only `transformprocessor`.

otelopscol/Makefile
franciscovalentecastro: Ok! Now i understand a bit more the problem space. Yes, having `go build` build the C sources and the CGO bindings would be ideal. It would remove most of the ways that `CGO` can become cumbersome moving forward. This requires further discussion.

[Current] I've explored extensively this part of the task and i have landed currently in a middle ground :
- Building and linking a static library `libonigmo.a`. Using `MingGW` for Windows since CGO doesn't seem to work with MSVC.
- Use `go-onigmo` and adding `ldflags/cflags` to the `go build` command.
- The resulting binary has the `libonigmo` library statically linked.
- I've tried putting all the sources on the same "folder" and tried to build (c sources too) everything with `go build`, but haven't achieved it yet (writing our own go binding would help a lot).

[Followup PR] How to improve this ?
- Use `oniguruma` instead of `onigmo`. Build tools for `oniguruma` are better (recently) maintained.
- Write our own `go-oniguruma` binding. Define exactly how our build environment should work.
- I propose to do this on a followup PR, because this will become an extensive update.

processor/agenttransformprocessor/internal/ottlfuncs/func_extract_patterns_ruby_regex.go
franciscovalentecastro: I'll explore how OTTL expects a return value for this case.

otelopscol/Makefile
franciscovalentecastro: Update :

It's indeed doable to only use `go build` without having to build `onigmo/oniguruma` separately and provide the static library. I did a "frankenstein" 🧟  combination, see https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/350, of the following to repos :

- https://github.com/kkos/oniguruma
- https://github.com/go-enry/go-oniguruma

Most of the changes required were to change `oniguruma` C code in a way that worked with `go build` (e.g. everything in the same folder).

franciscovalentecastro:
The result of the development in this PR became the new PR : https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/350. I'll close this PR in favour of that new implementation.

PR #286, opened 2025-04-23
Restructure repository for OTel version update

braydonk:
The structure for components in this repo had a fundamental issue. Once we introduced components that were specific to Google-Built OpenTelemetry Collector, we had the issue of otelopscol components targeting an older version of OpenTelemetry. We don't want to force all components to be on the same OTel version; when targeting different distributions, the components need to be at whichever version that distribution is on.

This PR restructures the component directories to separate them into two separate directories, one for each distribution the component could be targeting. The Makefile structure is adjusted around this fact. The new process for upating OTel components for a distribution is ~~to go into the Makefile for that distribution's components folder and update the version variables~~ update the distrogen spec with the new intended OTel versions and then in the main directory call the associated target to update those components.

This PR also does the OTel version updates for Google-Built OpenTelemetry Collector to prove that the restructure worked effectively.

Updates were needed to the client auth extension based on: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38451 https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38601


Makefile
```diff
@@ -26,34 +26,19 @@ presubmit: checklicense misspell lint compare-all
...
+update-google-otel-components: export OTEL_CONTRIB_VERSION := v$(shell $(GOOGLE_OTEL_SPEC_QUERY) opentelemetry_contrib_version)
+update-google-otel-components: install-tools
+	cd components/google-built-opentelemetry-collector && PATH="$(TOOLS_DIR):${PATH}" $(MAKE) update-components
+
+OTELOPSCOL_SPEC_QUERY = go run ./cmd/distrogen -spec specs/otelopscol.yaml -query
+.PHONY: update-otelopscol-components
+update-otelopscol-components: export OTEL_VERSION := v$(shell $(OTELOPSCOL_SPEC_QUERY) opentelemetry_version)
+update-otelopscol-components: export OTEL_CONTRIB_VERSION := v$(shell $(OTELOPSCOL_SPEC_QUERY) opentelemetry_contrib_version)
+update-otelopscol-components: install-tools
+	cd components/otelopscol && PATH="$(TOOLS_DIR):${PATH}" $(MAKE) update-components
```
jinghan-ma, line 41: Very nit picky, but it takes a bit of reading to realize that these two do the exact same thing, just on different directories. If we expect this to remain true in the future, maybe combining the definitions would make it our intentions more clear?

Something like:
```
.PHONY: update-google-otel-components update-otelopscol-components


update-google-otel-components: SPEC_FILE := specs/google-built-opentelemetry-collector.yaml
update-google-otel-components: COMPONENT_DIR := components/google-built-opentelemetry-collector

update-otelopscol-components: SPEC_FILE := specs/otelopscol.yaml
update-otelopscol-components: COMPONENT_DIR := components/otelopscol

update-google-otel-components update-otelopscol-components: DISTROGEN_QUERY := go run ./cmd/distrogen -spec $(SPEC_FILE) -query
update-google-otel-components update-otelopscol-components: export OTEL_VERSION := v$(shell $(DISTROGEN_QUERY) opentelemetry_version)
update-google-otel-components update-otelopscol-components: export OTEL_CONTRIB_VERSION := v$(shell $(DISTROGEN_QUERY) opentelemetry_contrib_version)
update-google-otel-components update-otelopscol-components: install-tools
    cd $(COMPONENT_DIR) && PATH="$(TOOLS_DIR):${PATH}" $(MAKE) update-components
```

I don't have strong opinions either way.

cmd/distrogen/distribution.go
```diff
@@ -295,6 +331,15 @@ func getGeneratedFilesInDir(dir string) (map[string]*generatedFile, error) {
 			return nil
 		}
 
+		// Don't include .tools directory in comparison.
+		if strings.Contains(path, ".tools") {
```
jinghan-ma, line 335: Another super nitpick.

Since we already exclude all directories in the previous if statement. I think checking for `/.tools/` in the path should still exclude contents in the `.tools` directory while excluding false positives like files/directories that contains the `.tools` substring like `google.otel.tools` (although it's super unlikely)

make/component_dir.mk
```diff
@@ -0,0 +1,43 @@
+TOOLS_DIR = $(PWD)/.tools
+MDATAGEN_BIN = $(TOOLS_DIR)/mdatagen
+
+.PHONY: update-otel-components
+update-components: export OTEL_VERSION := $(OTEL_VERSION)
+update-components: export OTEL_CONTRIB_VERSION := $(OTEL_CONTRIB_VERSION)
+update-components: update-deps tidy-components update-mdatagen generate-components
```
jinghan-ma, line 7: Are these intended to be `update-components` or `update-otel-components`?

jinghan-ma (commented):
It would also be nice to add a short section in the readme to document the process for changing the target version (e.g. change version in .yaml spec, run `update-google-otel-components`, etc.)

XuechunHou (approved):

Makefile
braydonk: Applied suggestion in [29ed4cf](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/286/commits/29ed4cf28bbab2e893242f7bc87f66e05437845f)

cmd/distrogen/distribution.go
braydonk: Excluding all directories just excludes the individual entry for that directory; `filepath.WalkDir` will still recursively walk the subdirectory even if we ignore the specific directory entry it presented us. So the check is still necessary.

Applied suggestion to make the check stricter in [8dc4905](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/286/commits/8dc4905b87df04316c5577fd633d55b8ddad9e64)

braydonk:
> It would also be nice to add a short section in the readme to document the process for changing the target version (e.g. change version in .yaml spec, run update-google-otel-components, etc.)

Added instructions in [397bae8](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/286/commits/397bae82a7ce1977590ed811d633b634329d0803)

LujieDuan (approved):
LGTM!

PR #292, opened 2025-04-29
Add tests to CI

braydonk:
This PR adds the necessary testing to build distros and test components so that PRs can be blocked upon a test failure or an issue with distro building.


.github/workflows/presubmit.yaml
```diff
@@ -44,4 +47,37 @@ jobs:
...
+        with:
+          submodules: true
+          fetch-depth: 0
+          lfs: true
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ matrix.go }}
+          cache: true
+      - name: Test Components
+        run: make test-${{matrix.distro}}-components
```
LujieDuan, line 69: Looks like the target is not running https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/actions/runs/14735595188/job/41360745228 - probably need to do `make workspace` first so that `go list` will list modules

braydonk: Done in [73d858c](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/292/commits/73d858ce936698ad1db41e43388df9594aa0ba9a)

LujieDuan (approved):
LGTM!

Makefile
```diff
@@ -40,6 +38,15 @@ update-google-otel-components update-otelopscol-components: export OTEL_CONTRIB_
...
 	cd $(COMPONENT_DIR) && PATH="$(TOOLS_DIR):${PATH}" $(MAKE) update-components
 
+.PHONY: test-google-otel-components test-otelopscol-components
+
+test-google-otel-components: COMPONENT_DIR := components/google-built-opentelemetry-collector
+
+test-otelopscol-components: COMPONENT_DIR := components/otelopscol
+
+test-google-otel-components test-otelopscol-components:
+	cd $(COMPONENT_DIR) && $(MAKE) test-components
```
LujieDuan, line 48: Looks like the tests on components are still not running: https://screenshot.googleplex.com/5GpubrWH7DQMvfw

Those `test-` targets here are calling the `target-all-modules` target in the components directory subfolders, and `go.work` is not executed before them.

braydonk: Yep thanks for catching that too. Done in [39480b3](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/292/commits/39480b345711d5d773aca09fd5b53faeb56dbd87)

LujieDuan (approved):
LGTM!

PR #294, opened 2025-04-30
Add cross-platform container build

braydonk:
This PR adds a new recipe for cross-platform container builds. It makes some changes to the Dockerfile.build and Makefile templates to make the process easier.


ridwanmsharif (approved):

cmd/distrogen/templates/Dockerfile.build.go.tmpl
```diff
@@ -14,15 +14,13 @@ ARG PROJECT_ROOT="."
 ARG CERT_CONTAINER="alpine:3"
 {{ if eq .BuildContainer "ubuntu" -}}
 ARG BUILD_CONTAINER="ubuntu:24.04"
-
-FROM ${BUILD_CONTAINER} AS build
+FROM --platform=${BUILDPLATFORM:-linux/amd64} ${BUILD_CONTAINER} AS build
```
igorpeshansky, line 17: [Optional] Is there any way this can be composed from `BUILDARCH`/`BUILDOS`? Alternatively, can `BUILDARCH`/`BUILDOS` be parsed out from `BUILDPLATFORM`?

braydonk: As far as I can tell, this is the intended usage for `BUILDPLATFORM` and `TARGETPLATFORM`: https://docs.docker.com/build/building/multi-platform/#cross-compilation

So there probably isn't much benefit to using `BUILDARCH` and `BUILDOS`, as they'll all be set anyway.

igorpeshansky: Right, if Docker sets them for you, there's no reason to optimize it.

PR #322, opened 2025-05-28
Makefile cleanups

braydonk:
Doing some Makefile cleanups that have been needed for a bit.

* Changed the `:=` evaluators for target variables that call shell programs. This was causing make tab complete to lag.
* Fixed the `tag-repo` target to use `distrogen -query`
* Fixed `collector-package-list` target in distro Makefile template

### DIFF
--- a/otelopscol/Makefile
+++ b/otelopscol/Makefile
@@ -123,9 +123,9 @@ $(COLLECTOR_BINARY_NAME): ocb-generate
 .PHONY: collector-package-list
 collector-package-list: ocb-generate
 	cd $(COLLECTOR_BUILD_TEMP_DIR) && \
-		GOWORK=off go list -deps -f '{{ .ImportPath }} {{ .Module }}' ./... > ../packages.txt
-	grep -v '<nil>' packages.txt | sort -u > packages.txt
-	rm -rf $(COLLECTOR_BUILD_TEMP_DIR)
+		GOWORK=off go list -deps -f '{{ .ImportPath }} {{ .Module }}' ./... |\
+		grep -v '<nil>' |\
+		sort -u > ../packages.txt
 
 # Delete any existing built collector binary.
 .PHONY: clean-collector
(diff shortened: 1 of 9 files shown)


otelopscol/Makefile
```diff
@@ -123,9 +123,9 @@ $(COLLECTOR_BINARY_NAME): ocb-generate
 .PHONY: collector-package-list
 collector-package-list: ocb-generate
 	cd $(COLLECTOR_BUILD_TEMP_DIR) && \
-		GOWORK=off go list -deps -f '{{ .ImportPath }} {{ .Module }}' ./... > ../packages.txt
-	grep -v '<nil>' packages.txt | sort -u > packages.txt
-	rm -rf $(COLLECTOR_BUILD_TEMP_DIR)
+		GOWORK=off go list -deps -f '{{ .ImportPath }} {{ .Module }}' ./... |\
+		grep -v '<nil>' |\
+		sort -u > ../packages.txt
```
franciscovalentecastro, line 128: Is it possible to move `collector-package-list` (and possibly other targets) to a top-level `Makefile.common` that could be include here ?

I've recently found that `opentelemetry-collector-contrib` does this in every component which is very convenient.

braydonk: Anything that is within the generated distribution folder has to come from the `distrogen` template in `cmd/distrogen/templates`. Any of these templates are designed to be completely standalone from where they're generated, so they won't share any code.

I do use [reusable Make code](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/tree/master/make) that is [imported by our components](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/components/google-built-opentelemetry-collector/Makefile) since the components in our repo are specific to here.

franciscovalentecastro: Thanks for the explanation!

PR #328, opened 2025-05-29
Hardcode windows bin path as the glob doesn't seem to be working

martijnvans:
http://sponge2/6f1af738-531d-4534-86c2-eec84ff8ce7d

### DIFF
--- a/kokoro/scripts/build/sign_windows_binaries.bat
+++ b/kokoro/scripts/build/sign_windows_binaries.bat
@@ -12,7 +12,7 @@
 @REM See the License for the specific language governing permissions and
 @REM limitations under the License.
 
-ksigntool sign GOOGLE_EXTERNAL /v /debug /t http://timestamp.digicert.com "%KOKORO_GFILE_DIR%"\dist\*\*.exe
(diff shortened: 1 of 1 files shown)


kokoro/scripts/build/sign_windows_binaries.bat
```diff
@@ -12,7 +12,7 @@
 @REM See the License for the specific language governing permissions and
 @REM limitations under the License.
 
-ksigntool sign GOOGLE_EXTERNAL /v /debug /t http://timestamp.digicert.com "%KOKORO_GFILE_DIR%"\dist\*\*.exe
+ksigntool sign GOOGLE_EXTERNAL /v /debug /t http://timestamp.digicert.com "%KOKORO_GFILE_DIR%"\dist\otelcol-google-windows_windows_amd64_v1\*.exe
```
braydonk, line 15: I'm not sure what glob syntax this uses, might be `**` for dirs like in gitignore and similar glob formats. Hardcoding this seems fine though I don't think there's any other exes we want to sign.

braydonk (approved):

kokoro/scripts/build/sign_windows_binaries.bat
igorpeshansky: According to https://stackoverflow.com/a/51176611:
> Wildcards can only be used in the _last_ element of a path…

Maybe we need the `for` loop syntax here?

PR #342, opened 2025-06-16
Updating example docs

avilevy18:
Updating example docs config to compatible with OTel version 1.28.0

### DIFF
--- a/google-built-opentelemetry-collector/docs/examples/deployment/cos/cloud-init-config-example.yaml
+++ b/google-built-opentelemetry-collector/docs/examples/deployment/cos/cloud-init-config-example.yaml
@@ -23,20 +23,6 @@ write_files:
               - https://*
             endpoint: localhost:4318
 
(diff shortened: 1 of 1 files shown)


google-built-opentelemetry-collector/docs/examples/deployment/cos/cloud-init-config-example.yaml
```diff
@@ -148,4 +124,9 @@ write_files:
           - googlecloud
       telemetry:
         metrics:
-          address: localhost:8888
+          readers:
```
braydonk, line 127: We should add a descriptive comment to this section similar to the descriptive comments on all the sections above.

braydonk (commented):
The update should be done by applying the change in `templates/google-built-opentelemetry-collector/docs` and `make regen-google-built-otel`. Additionally, the same config change needs to be applied to `config-standard.yaml.go.tmpl` along with this one.

PR #345, opened 2025-06-19
Access AR repositories via GOPROXY in GBOC

XuechunHou:
1. Switched to a different docker container image that has gcloud  for GBOC.   
Airlock requires referencing a docker image by [digest](https://screenshot.googleplex.com/3V4byzjgEbNZs5K). 
Found the docker image on Airlock: https://screenshot.googleplex.com/F367q7Lta8VS44V

3. Created a separate set of docker templates and make file templates for Assured-GBOC to pull packages from AOSS AR repo via `GOPROXY` 

See a successful docker build on Louhi via Kokoro: https://fusion2.corp.google.com/invocations/30312fd9-8944-4433-8735-d31bb52341f3/targets/stackdriver_agents%2Fotelcol-google%2Fexperiment_aoss_build_image;config=default/log

Note: We need this change regardless of pointing to Airlock repos or AOSS repos.


templates/google-built-opentelemetry-collector/Makefile.pull_modules_from_ar_repo.go.tmpl
```diff
@@ -0,0 +1,206 @@
+# NOTE: File generated by distrogen. Do not manually edit.
```
braydonk, line 1: Take a look at what we do in the `goopack.mk` template: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/master/templates/google-built-opentelemetry-collector/goopack.mk.go.tmpl

You can `include` the main Makefile so that you have access to everything you need from the main one and just add the new stuff you need.

I would recommend you change the name of this to `aoss.mk.go.tmpl` to follow the same pattern we're already using for supplemental Makefiles, and then make the contents the following:

```make
include ./Makefile

.PHONY: aoss-build
aoss-build: $(AR_AUTH_BIN)
	$(AR_AUTH_BIN) add-locations --locations=us && \
	$(AR_AUTH_BIN) refresh && \
	$(MAKE) build

AR_AUTH_BIN = $(TOOLS_DIR)/auth

$(AR_AUTH_BIN): $(GO_BIN)
	@{ \
	set -e ;\
	mkdir -p $(TOOLS_DIR) ;\
	echo "Installing AR Auth tools at $(TOOLS_DIR)" ;\
	GOBIN=$(TOOLS_DIR) CGO_ENABLED=0 $(GO_BIN) install -trimpath -ldflags="-s -w" github.com/GoogleCloudPlatform/artifact-registry-go-tools/cmd/auth@v0.4.0 ;\
	} 
```

(I wrote this off the cuff in the comment so you'll probably need to tweak it locally if things don't work)

google-built-opentelemetry-collector/Makefile
```diff
@@ -24,7 +24,7 @@ GO_BIN_DIR ?= $(TOOLS_DIR)/go/bin
 GO_BIN ?= $(GO_BIN_DIR)/go
 SET_GO_BIN_PATH ?= PATH="$(GO_BIN_DIR):$$PATH"
 
-USE_GO_PROXY ?= 
+USE_GO_PROXY ?= "https://us-go.pkg.dev/access-aoss/assuredoss-go"
```
braydonk: What's happening here? This is intended to be formatted in directly from the value in the spec but this looks like a manual change to the generated Makefile.

braydonk: Disregard; GitHub UI didn't refresh.

templates/google-built-opentelemetry-collector/Dockerfile.image_with_gcloud.go.tmpl
```diff
@@ -0,0 +1,15 @@
+# NOTE: File generated by distrogen. Do not manually edit.
```
braydonk, line 1: This does not appear to serve a purpose and can probably be deleted.

braydonk: Sorry I should have clarified; I meant the entire existence of this file. I think the whole template can be deleted.

braydonk (approved):

PR #350, opened 2025-06-25
[Draft] Create `oniguruma` go bindings and use `go build`.

franciscovalentecastro:


quentinmit (commented):
I think the OTTL function (and library) should not live inside `transformprocessor`, since we're also going to be adding these functions to `filterprocessor` Maybe just move the `internal` directory up to the top-level?

otelopscol/manifest.yaml
```diff
@@ -8,78 +8,77 @@ dist:
   output_path: ./_build
```
braydonk, line 8: Looks like this was edited manually. This will need to be properly generated by `distrogen` before merge. The process will be:

* Change the `opentelemetry_version` fields in `specs/otelopscol.yaml` 
* Add the local `transformprocessor` to `components/otelopscol/registry.yaml`

Once you regenerate you should have the same setup as what you've manually done here.

otelopscol/build.ps1
```diff
@@ -29,10 +29,21 @@ param (
     [string]$outDir = "."
 )
 
+# Store PATH
```
braydonk, line 32: Looks like this was edited manually. This will need to be properly generated by distrogen before merge.

You'll need to make this updates to `templates/otelopscol/build.ps1.go.tmpl`. Once you regenerate you should have the same as what you have here.

otelopscol/manifest.yaml
franciscovalentecastro: Thank you for the comment! Updated `specs/otelopscol.yaml`,  `components/otelopscol/registry.yaml` and ran distrogen. Done :)

otelopscol/build.ps1
franciscovalentecastro: Sounds good! I updated `build.ps1.go.tmpl` and ran distrogen. Done!

third_party/Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: generate-third-party-deps
+generate-third-party-deps:
+	git clone git@github.com:go-enry/go-oniguruma.git
```
quentinmit, line 3: I wouldn't clone the Git repo inside the parent repo like this; you're removing the .git and .github directories, but that's going to be very fragile.

If you want to offer this, I would put it in a temporary directory.
```
third_party/go-oniguruma:
  DIR=$$(mktemp -d) \
  cd "$DIR" && \
  git clone --branch v1.2.1 git@github.com:go-enry/go-oniguruma.git go-oniguruma && \
  cd go-oniguruma && \
  git apply $(abspath "go-oniguruma.patch") && \
  rm -r "$(abspath $@)" && \
  cp -a "$DIR/go-oniguruma" "$(abspath $@)"
```

(and repeat with a separate rule for oniguruma.git)

third_party/go-oniguruma.patch
```diff
@@ -0,0 +1,90 @@
...
+ #include <stdlib.h>
+-#include <oniguruma.h>
+ #include "chelper.h"
+ */
+ import "C"
+@@ -14,6 +11,7 @@ import (
+ 	"errors"
+ 	"fmt"
+ 	"io"
++	_ "oniguruma"
```
quentinmit, line 56: Why not use an absolute package path here instead of something you're `replace`-ing?

third_party/oniguruma.patch
```diff
@@ -0,0 +1,400 @@
+diff --git a/.gitignore b/.gitignore
+deleted file mode 100644
+index 71ffb4c..0000000
+--- a/.gitignore
```
quentinmit, line 4: Does the .gitignore cause any problems if you leave it in place?

third_party/Makefile
franciscovalentecastro: Thank you for this suggestion! This is better, though i'm still curious why cloning in a temp folder and then copying works better. It does work without having to remove the `.git` folder, but why does this work ? 🤔 

I also created `update_go-oniguruma.patch` and `update_oniguruma.patch` rules with this setup.

components/otelopscol/registry.yaml
```diff
@@ -41,3 +41,7 @@ processors:
   normalizesums:
     gomod: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/components/otelopscol/processor/normalizesumsprocessor v0.0.0
     path: "../components/otelopscol/processor/normalizesumsprocessor"
+  transform:
+    gomod: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/components/otelopscol/processor/transformprocessor v0.0.0
+    path: "../components/otelopscol/processor/transformprocessor"
+    docs_url: https://www.github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor/README.md
```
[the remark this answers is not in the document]
franciscovalentecastro: Done!

specs/otelopscol.yaml
```diff
@@ -84,3 +84,11 @@ components:
...
     - file
     - env
+
+replaces:
+  - from: oniguruma
+    to: ../../third_party/oniguruma/src
+    reason: Modified version of third party library.
+  - from: github.com/go-enry/go-oniguruma v1.2.1
+    to: ../../third_party/go-oniguruma/
+    reason: Modified version of third party library.
```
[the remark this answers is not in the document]
franciscovalentecastro: I don't think this is possible (@braydonk can corroborate further) :

- AFAIU : The OCB build happens in a separate `_build` folder, so go won't be able to resolve automatically the dependencies when developing locally. We could potentially Github tag this "third_party" forks so they could be resolved remotely, but i think this brings other issues.

- The alternative to use a `go.work` workspace doesn't work with the OCB build since it uses the option `GoWork=off`.

third_party/go-oniguruma.patch
franciscovalentecastro: I haven't found a concrete reason so far to give `oniguruma` an absolute package path. I have a couple of reasons not to do it : 

- Considering the use of `tags` so remote dependencies with abosulte paths would work would required that the updates to `otelopscol` be aligned with `google-built-opentelemetry-collector` releases, which may cause issues.

- We still need use a `replace` to build `otelopscol` so it would work for local development. This because Go doesn't honor `replace`s from other packages `go.mod` (https://go.dev/wiki/Modules#gomod).

- Due to the `third_party` folder structure, resolving import paths doesn't work because Go requires subpackages to be in the same `folder` tree.

- Without an absolute path we can avoid other repo's having dependencies on this modified version of the packages in `third_party` (`oniguruma`, `go-oniguruma`).

CC @braydonk

components/otelopscol/processor/transformprocessor/func_extract_patterns_ruby_regex.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+
+	rubex "github.com/go-enry/go-oniguruma"
+)
+
+type ExtractPatternsRubyRegexArguments[K any] struct {
+	Target  ottl.StringGetter[K]
```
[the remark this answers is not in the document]
quentinmit: We should match `ExtractPatterns`. My thought was that in edge cases users might want to apply regexes to other types (e.g. using a regex to truncate a float, or extract digits from an int), and I think a `StringLikeGetter` would allow that. But if `ExtractPatterns` doesn't do it I agree we also shouldn't do it here.

third_party/go-oniguruma.patch
quentinmit: > I haven't found a concrete reason so far to give `oniguruma` an absolute package path. I have a couple of reasons not to do it :
> 
> * Considering the use of `tags` so remote dependencies with abosulte paths would work doesn't seem feasible. It would require that updates to `otelopscol` be aligned with `google-built-opentelemetry-collector` releases, which may cause issues.

This pertains to eliminating `replace`, no? That seems orthogonal to whether we use an absolute path or not.

> * Due to the `third_party` folder structure, resolving local import paths doesn't work because Go requires subpackages to be in the same `folder` tree (https://go.dev/wiki/Modules#gomod).

Why is it a subpackage? You currently have two packages defined in this PR:
- `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/third_party/oniguruma`
- `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/third_party/go-oniguruma`
so neither package is a subpackage of the other. You're using `replace` to import `../oniguruma` as `oniguruma`, but you could just as easily use `replace` to import `../oniguruma` as `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/third_party/oniguruma` and then you wouldn't be inventing any new package paths.

> * We still need to use a `replace` to build `otelopscol`, this because Go doesn't honor `replace`s from other packages `go.mod` (https://go.dev/wiki/Modules#gomod).

I guess I need to look more at `ocb`, but (assuming we need separate modules in the first place) it looks like a `use` directive in a module-specific workspace is probably the best way to pull in other modules in the same repo, instead of `replace` directives.

> * Without an "absolute path" we can avoid other repo's having dependencies on this modified version of the packages in `third_party` (`oniguruma`, `go-oniguruma`).

How we import the package has no bearing on how other people import the package. With the current state of the PR, I believe what you're calling `oniguruma` *is* accessible at `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/third_party/oniguruma`, regardless of how you import it from `go-oniguruma`.

> CC @braydonk

Can the three of us sync on Monday to talk though the constraints here?

third_party/Makefile
quentinmit: `git`'s behavior inside a repository is partially controlled by that repository. Cloning in a temporary directory ensures that `git` is executing from a clean state, and neither repository is polluting the other.

third_party/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	echo "$(abspath $@)" && \
+	cd $$DIR && \
+	git clone --branch v6.9.10 git@github.com:kkos/oniguruma.git oniguruma && \
+	cd oniguruma && \
+	git apply $(abspath oniguruma.patch) && \
+	rm -rf "$(abspath $@)" && \
+	cp -a $$DIR"/oniguruma" "$(abspath $@)"
+
+.PHONY: update_go-oniguruma.patch
+update_go-oniguruma.patch:
```
quentinmit, line 22: Why did you call this `update_go-oniguruma.patch` instead of just calling it `go-oniguruma.patch` since that's what it creates? These aren't phony rules; they're actually rules to make a file.

third_party/Makefile
```diff
@@ -0,0 +1,39 @@
...
+	cd go-oniguruma && \
+	git add . && \
+	git diff --staged > $(abspath go-oniguruma.patch)
```
quentinmit, line 25: I don't think this calculates the diff that you want. This calculates the diff between the working copy's go-oniguruma and the checked-in go-oniguruma. But you want to calculate the diff between the working copy's go-oniguruma and the *upstream* go-oniguruma.

third_party/Makefile
franciscovalentecastro: Thanks for the explanation!

third_party/Makefile
franciscovalentecastro: Updated the command so now it compares directly with the remote tag which we use to clone :
```
	git fetch --tags origin && \
	git add . && \
	git diff --staged v1.2.1 > $(abspath go-oniguruma.patch)
```

Done.

third_party/go-oniguruma.patch
franciscovalentecastro: Update :

I changed the module names to follow the expected absolute paths :  
- `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/internal/ottlfuncs`
- `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/third_party/go-oniguruma`
- `github.com/GoogleCloudPlatform/opentelemetry-operations-collector/third_party/oniguruma/src`

The current implemented solution still uses `replace` to refer locally to modules. After making a release of this repo, this modules would be referenced externally using tag like `v0.130.0`.

specs/otelopscol.yaml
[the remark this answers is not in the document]
franciscovalentecastro: Update : As we discussed offline with @braydonk, since the repo is organized with every component as a separate module (transformprocessor/go.mod), currently we cannot avoid using `replace` locally. Some details :

- When this changes are released & tagged, this modules would be referenced externally using tag like `v0.130.0`.
- I've created a followup task to consider the following changes : 
  - Only one `opentelemetry-operations-module` so all components can be referenced as sub-folder packages.
  - Use a committed `go.work` file with `use` directive that would be used instead of `replace`.

third_party/Makefile
franciscovalentecastro: Update : As pointed out by @quentinmit, this command was relying on it still being a Git repo (having a `.git`). The assumption was that everything under `third_party` should not be a Git repo to avoid polluting the main repo. I fixed both things : 
- The copy command to `third_party` uses `$$DIR"/go-oniguruma"/*` to avoid copying any hidden file (e.g. `.git`).
- The `go-oniguruma.patch` rule now copies all files in `go-oniguruma/*` folder to a freshly cloned repo and executes "git diff" into `go-oniguruma.patch`.

https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/3d39dcbed125ccd45522570fbc9cd146d9bf6f76/third_party/Makefile#L4-L53

specs/otelopscol.yaml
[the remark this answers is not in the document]
franciscovalentecastro: Update - Now the `transformprocessor` relies on only the top-level module to find its dependencies (i removed the `internal/ottlfuncs` and `third_party` modules) : 
```
replaces:
  - from: github.com/GoogleCloudPlatform/opentelemetry-operations-collector
    to: ../../
    reason: Reference root module of repository to find dependency packages.
```

quentinmit (approved):

PR #370, opened 2025-07-22
Added a custom ottl function: `ToValues`

XuechunHou:
`ToValues` takes an array of `pcommon.Map` and converts it to an array of values of the maps. 

Fixes: b/424476367


internal/ottlfuncs/func_to_values_test.go
```diff
@@ -0,0 +1,66 @@
...
+						return m, nil
+					},
+				},
+				ottl.StandardPMapGetter[any]{
+					Getter: func(_ context.Context, _ any) (any, error) {
+						m := pcommon.NewMap()
+						m.PutStr("param2", "stopped")
+						return m, nil
+					},
+				},
```
quentinmit, line 50: Swap the order of these maps so you can make sure it properly preserves the original order (i.e. `param2` and then `param1`)

internal/ottlfuncs/func_to_values.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"errors"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+)
+
+type ToValuesArguments[K any] struct {
+	Target []ottl.PMapGetter[K]
```
quentinmit, line 27: Does a slice get handled correctly by the ottl package? I would expect to need a `PSliceGetter` here.

internal/ottlfuncs/func_to_values.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil, errors.New("ToValuesFactory args must be of type *ToValuesArguments[K]")
+	}
+
+	return toValues(args.Target)
+}
+
+func toValues[K any](target ottl.PSliceGetter[K]) (ottl.ExprFunc[K], error) {
+
+	return func(ctx context.Context, tCtx K) (any, error) {
+		res := []any{}
```
quentinmit, line 47: Why bother with the intermediate result here and the extra conversions to/from `pcommon.Slice`?

```go
res := pcommon.NewSlice()
...
  for _, v := range v.Map().All() {
    v.CopyTo(res.AppendEmpty())
  }
...
return res, nil
```

quentinmit (approved):

PR #372, opened 2025-07-22
Distrogen update

avilevy18:
Updating distrogen to add full project generation. It assumes that there will always be one distribution per project.

### DIFF
--- a/cmd/distrogen/distribution.go
+++ b/cmd/distrogen/distribution.go
@@ -65,9 +65,9 @@ type DistributionSpec struct {
 	GoProxy                     string                  `yaml:"go_proxy,omitempty"`
 
 	// CollectorCGO determines whether the Collector will be built with CGO.
-	CollectorCGO bool `yaml:"collector_cgo,omitempty"`
-	ComponentModuleBase         string                  `yaml:"component_module_base"`
-	DistrogenVersion            string                  `yaml:"distrogen_version"`
+	CollectorCGO        bool   `yaml:"collector_cgo,omitempty"`
+	ComponentModuleBase string `yaml:"component_module_base"`
+	DistrogenVersion    string `yaml:"distrogen_version"`
 }
 
 // RenderGoMajorVersion will parse the GoVersion in the spec and return a version without a patch
@@ -185,7 +185,7 @@ func NewDistributionGenerator(spec *DistributionSpec, registry *Registry, forceG
 		// -rw-r--r--
 		FileMode: DefaultFileMode,
 	}
-	d.GenerateDirName = spec.BinaryName
+	d.GenerateDirName = spec.Name
 
 	if !forceGenerate {
 		specCache, err := yamlUnmarshalFromFile[DistributionSpec](filepath.Join(d.GenerateDirName, "spec.yaml"))
--- a/cmd/distrogen/testdata/generator/distribution/build_container/golden/spec.yaml
+++ b/cmd/distrogen/testdata/generator/distribution/build_container/golden/spec.yaml
@@ -30,3 +30,5 @@ components:
 feature_gates:
     - exporter.googlemanagedprometheus.intToDouble
 collector_cgo: true
+component_module_base: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/basic
+distrogen_version: v0.130.1-0.20250724183747-261516e1bff1


cmd/distrogen/distribution.go
```diff
@@ -60,8 +61,19 @@ type DistributionSpec struct {
...
 	Replaces                    ComponentReplaces       `yaml:"replaces,omitempty"`
 	CustomValues                map[string]any          `yaml:"custom_values,omitempty"`
-	FeatureGates                FeatureGates            `yaml:"feature_gates"`
+	FeatureGates                FeatureGates            `yaml:"feature_gates,omitempty"`
 	GoProxy                     string                  `yaml:"go_proxy,omitempty"`
+	ComponentModuleBase         string                  `yaml:"component_module_base"`
+	DistrogenVersion            string                  `yaml:"distrogen_version"`
+}
+
+// GetGoMajorVersion will parse the GoVersion in the spec and return a version without a patch
```
braydonk, line 70: I have a light convention which I did not write down or tell anyone which is that functions intended to be used in templates should be named `RenderX`. Could we change the name of this to `RenderGoMajorVersion`?

cmd/distrogen/component.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (g *ComponentGenerator) Generate() error {
+	componentTemplates, err := GetIndividualComponentTemplateSet(g, g.FileMode)
+	if err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(g.Path, g.FileMode); err != nil {
+		logger.Debug("generator", "generator", *g)
```
braydonk, line 70: These debug logs can probably be removed when you cleanup and make the PR ready for full review.

cmd/distrogen/distrogen.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+import (
+	"io/fs"
+	"os"
+)
+
+type Distrogen struct {
```
braydonk, line 8: Since this is nominally a part of a "project" (this is essentially a "distrogen project folder") is it possible to fold this functionality into the `ProjectGenerator`?

cmd/distrogen/project.go
```diff
@@ -0,0 +1,86 @@
...
+	if err := GenerateTemplateSet("make", makeTemplates); err != nil {
+		return err
+	}
+	if err := GenerateTemplateSet(".", projectTemplates); err != nil {
+		return err
+	}
+
+	return nil
+
+	// renderFail:
```
braydonk, line 74: This can be safely removed.

cmd/distrogen/templates/project/Makefile
```diff
@@ -0,0 +1,173 @@
...
+###################
+# Distro Generation
+###################
+
+.PHONY: gen
+gen: distrogen
+	@$(GEN_OTEL)
+
+.PHONY: regen
+regen: $(DISTROGEN_BIN)
```
braydonk, line 53: These should all depend on `distrogen` like `gen` does.

cmd/distrogen/templates/project/Makefile
```diff
@@ -0,0 +1,173 @@
...
+	@$(GEN_OTEL) --force
+
+regen-v: $(DISTROGEN_BIN)
+	@$(GEN_OTEL) --force -v
+
+GEN_OTEL = $(DISTROGEN_BIN) generate --spec $(SPEC_FILE) \
```
braydonk, line 59: Doesn't actually change anything but it's easier to read if this is above the gen targets.

cmd/distrogen/templates/project/Makefile
```diff
@@ -0,0 +1,173 @@
...
+	rm -f go.work
+	rm -f go.work.sum
+
+#######
+# Tools
+#######
+
+TOOLS_DIR = $(PWD)/.tools
+DISTROGEN_BIN ?= $(TOOLS_DIR)/distrogen
+GOLANGCI_LINT = $(TOOLS_DIR)/golangci-lint
```
braydonk, line 112: I believe we decided these were unnecessary?

cmd/distrogen/templates/project/Makefile
```diff
@@ -0,0 +1,173 @@
...
+	@mkdir -p $(TOOLS_DIR)
+
+TOOL_LIST ?= github.com/GoogleCloudPlatform/opentelemetry-operations-collector/cmd/distrogen@ec10f6e4fdf4b913eaa4b9be21715576e6e455c9
+
+.PHONY: install-distrogen
+install-distrogen: tools-dir
+	GOBIN=$(TOOLS_DIR) go install github.com/GoogleCloudPlatform/opentelemetry-operations-collector/cmd/distrogen@$(shell cat .distrogen/VERSION)
+
+.PHONY: lint
+lint:
```
braydonk, line 131: These tool targets can be removed.

cmd/distrogen/templates/project/Makefile
```diff
@@ -0,0 +1,173 @@
...
+# it gets very confused because the creation date of the .tools
+# directory is newer than the tools inside it.
+.PHONY: tools-dir
+tools-dir:
+	@mkdir -p $(TOOLS_DIR)
+
+TOOL_LIST ?= github.com/GoogleCloudPlatform/opentelemetry-operations-collector/cmd/distrogen@ec10f6e4fdf4b913eaa4b9be21715576e6e455c9
+
+.PHONY: install-distrogen
+install-distrogen: tools-dir
```
braydonk, line 127: I don't think this 3rd level of indirection is necessary; `distrogen` depending on `$(DISTROGEN_BIN)` and then this recipe being directly for the `$(DISTROGEN_BIN)` target seems like it should be fine.

braydonk (commented):
I ran a manual test and most things I tried seemed to work. I encountered one bug, which is that both a `distrogen` and `.distrogen` folder were generated.

I think it would be a good idea to put together project generation golden test cases, similar to what we have in `testdata/generator`, which are defined in `distribution_test.go`. You could also add full project generation test cases to check those goldens. There is certainly more we could do for testing, but I think at least having that would be a good start (and would catch the bug I encountered).

After that, the `scratch` folder can be removed and we can formally move this to in review.

cmd/distrogen/registry.go
```diff
@@ -140,13 +185,15 @@ type otelComponentVersion struct {
...
-	DocsURL       string      `yaml:"docs_url"`
+	Name string `yaml:"-"`
+
+	GoMod         *GoModuleID               `yaml:"gomod"`
+	Import        string                    `yaml:"import,omitempty"`
+	Path          string                    `yaml:"path,omitempty"`
+	Stable        bool                      `yaml:"stable,omitempty"`
+	StartRevision string                    `yaml:"start_revision,omitempty"`
+	DocsURL       string                    `yaml:"docs_url,omitempty"`
+	Releases      RegistryComponentReleases `yaml:"releases,omitempty"`
```
braydonk, line 196: idk if I added this or you did, either way we can remove it for now since I'm handling it in my PR.

jefferbrecht:
Please add a motivation to the PR description. What problem is this solving? It doesn't look like a trivial change and the uninformed reader is not going to understand what's happening here.

cmd/distrogen/component.go
```diff
@@ -0,0 +1,105 @@
...
+		registry.Path = registryPath
+	}
+
+	registry.Add(g.Type, &RegistryComponent{
+		GoMod: &GoModuleID{
+			URL:           g.ModuleURL,
+			AllowBlankTag: true,
+		},
+		Name: g.Name,
+		Path: "../" + g.Path,
```
ridwanmsharif, line 98: So we are relying on this structure then? Where the registry is found in `components/registry.yaml`

users of this tool will own this repo. We should be documenting the structure we expect them to maintain somwehere

ridwanmsharif (commented):
> Please add a motivation to the PR description. What problem is this solving? It doesn't look like a trivial change and the uninformed reader is not going to understand what's happening here.

+1 to Jeffs comment, I got pretty lost in this deep dark forest so consider this a partial review for now

cmd/distrogen/registry.go
jefferbrecht, line 1: Are there any docs explaining the need for a component registry?

If I'm wearing my component author hat, what kind of indexing does the registry provide that the built-in module/package management system in Go doesn't provide already? I'm already able to broadcast my component to the world with the tooling already available.

If I'm wearing my distro author hat, I can already identify components directly by their URI. What is the registry's indirection providing me?

cmd/distrogen/component.go
avilevy18: Agreed, lets have a different PR documenting everything about distrogen.

cmd/distrogen/components.go
```diff
@@ -0,0 +1,55 @@
+package main
```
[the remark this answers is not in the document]
avilevy18: renamed `components.go` to `components_registry_generator.go`

cmd/distrogen/testdata/generator/project/golden/spec.yaml
```diff
@@ -0,0 +1,51 @@
...
+display_name: Basic OTel
+module: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/google-built-opentelemetry-collector/cmd/distrogen/testdata/generator/project/golden
+version: 0.121.0
+description: "A basic distribution of the OpenTelemetry Collector"
+blurb: "A basic collector distro"
+opentelemetry_version: 0.121.0
+opentelemetry_stable_version: 1.27.0
+binary_name: otelcol-basic
+collector_cgo: false
+go_version: 1.24.0
```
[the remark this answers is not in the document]
braydonk: Go version used to build gets picked up by vuln scans, so we want to be able to control it down to the exact version. There might be a mechanism by which we can only specify the minor version and then have distrogen always pull latest Go, but not sure exactly how to do that. (Maybe downloading Go ends up working with only the minor version to pull the latest patch version of that Go branch? idk)

cmd/distrogen/templates/project/go.mod.go.tmpl
```diff
@@ -0,0 +1,4 @@
+module {{ .Spec.Module }}
+
+go {{ .Spec.RenderGoMajorVersion }}
```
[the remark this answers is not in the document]
braydonk: This is a convention of the project (and in fact we got ourselves into trouble in the Collector for specifying the patch version of Go in our `operations-go` modules). If you specify the version it makes Go's automatic toolchain resolution do some automatic downloading of the version that some random module asked for. It might be fine for our case but I was just thinking we should follow the convention of upstream on this.

ridwanmsharif (approved):
I'll let the two of you resolve the open threads, I think I'm okay with most of the answers given as is - we can iterate on the rest in follow up changes

PR #390, opened 2025-08-05
Remove `time/tzdata` and set go version to `1.24` in `transformprocessor`.

franciscovalentecastro:
Due to the mismatch of the Go version in the `transformprocessor` (set to `1.24.4`) and the Ops Agent `otelopscol` build Go version (`1.24`), there was "bizarre" failure that resulted in `otelopscol` not finding the `tzdata` (timezone database) when parsing timestamps.

### Summary
- If the the OS doesn't provide the `tzdata` package, one of the other alternatives for it to find it is from the Go installation (e.g. "time/tzdata") using to build the binary.
- If a different version of Go is used to run tests, it may look in the incorrect place (wrong Go version folder ) for the `timezone` database.

Context :
- https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/382
- https://github.com/GoogleCloudPlatform/ops-agent/pull/2012


jefferbrecht:
IMO we should still add `tzdata` as an explicit dependency inside the test container. Relying on keeping both golang versions in sync is going to be brittle, and should ultimately be unnecessary: different patch levels of golang _should_ be inter-compatible.

franciscovalentecastro:
Re @jefferbrecht :

> IMO we should still add `tzdata` as an explicit dependency inside the test container. Relying on keeping both golang versions in sync is going to be brittle, and should ultimately be unnecessary: different patch levels of golang _should_ be inter-compatible.

I agree, we should add `tzdata` as an explicit dependency. After discussing with @quentinmit yesterday, importing `time/tzdata` directly in the Go binary is not the best solution since `tzdata` keeps getting updated and it would be better for the Ops Agent to keep using the OS version (e.g. deb package).

I think this two steps would be a good solution : 
1. Add `tzdata` as a `Recommended` dependency in our `deb` and `rpm` packages.
2. Install directly in the Ops Agent build docker containers.

As mentioned in https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/390, i will do a followup PR to do this changes in the Ops Agent.

components/otelopscol/processor/transformprocessor/go.mod
jefferbrecht, line 1: If we're going to add `tzdata` to the build container to fix this downstream then we shouldn't need this change. If there's a separate goal to keep the versions of golang in sync then I'd rather upgrade at the other side than downgrade at this side.

jefferbrecht: Nevermind I misunderstood what I was looking at, 1.24 should be fine.

franciscovalentecastro: I think keeping the version Go `1.24` here enables the downstream Go version to be any of `1.24.X`, so it is more flexible.

PR #404, opened 2025-08-22
Create `IsMatchRubyRegex` and add it to `filterprocessor`.

franciscovalentecastro:
This PR creates an OTTL function called `IsMatchRubyRegex` which works as `IsMatch` [^1] , but uses the `oniguruma` regex engine as implemented in https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/350. Also, this function is added to the `filterprocessor` (and `transformprocessor`) as a custom OTTL function to be used later in the Ops Agent.

b/409351024

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#ismatch


internal/ottlfuncs/func_is_match_ruby_regex.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+)
+
+type IsMatchRubyRegexArguments[K any] struct {
+	Target  ottl.StringLikeGetter[K]
+	Pattern string
+}
+
+func NewIsMatchRubyRegexFactory[K any]() ottl.Factory[K] {
+	return ottl.NewFactory("IsMatchRubyRegex", &IsMatchRubyRegexArguments[K]{}, createIsMatchRubyRegexFunction[K])
```
quentinmit, line 33: This line can move into `LogFunctions()` and you don't need this as a separate function at all

PR #406, opened 2025-08-25
[googleservicecontrol] Send userAgent with SC grpc clients

LujieDuan:
Set userAgent option to GRPC requests and make sure the option is used by both the client types. 
[b/440553736](http://b/440553736)

### DIFF
--- a/components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter/factory.go
+++ b/components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter/factory.go
@@ -105,7 +105,7 @@ func createLogExporter(ctx context.Context, settings exporter.Settings, cfg comp
 	exp := NewLogsExporter(*oCfg, settings.Logger, *c, settings.TelemetrySettings)
 	return exporterhelper.NewLogs(ctx, settings, cfg, exp.ConsumeLogs,
 		exporterhelper.WithCapabilities(exp.Capabilities()),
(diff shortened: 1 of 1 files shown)


components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter/factory.go
```diff
@@ -132,10 +133,15 @@ func createMetricsExporter(ctx context.Context, settings exporter.Settings, cfg
 }
 
 func createClient(ctx context.Context, oCfg *Config, settings exporter.Settings) (*ServiceControlClient, error) {
-	opts := []grpc.DialOption{}
+	userAgent := fmt.Sprintf("%s/%s (%s/%s)",
+		settings.BuildInfo.Description, settings.BuildInfo.Version, runtime.GOOS, runtime.GOARCH)
+	opts := []grpc.DialOption{
+		grpc.WithUserAgent(userAgent),
+	}
+	useRawServiceControlClient := strings.TrimSpace(strings.ToLower(oCfg.UseRawServiceControlClient)) == "true"
```
jefferbrecht, line 141: Nit: why is `oCfg.UseRawServiceControlClient` a string and not a bool? If it does need to accept values other than true or false, can we check in `Validate()` that it conforms to the rules here (is lower case and has no surrounding whitespace)? It'd be better if we didn't have to trim and normalize the string at the point of check.

PR #437, opened 2025-10-10
[chore] Start tagging googleservicecontrolexporter

braydonk:
To allow external users to reference the googleservicecontrolexporter we need to tag the repo for it. This PR updates the tagging script to do this. Eventually this should be fully automated but this should unblock for now.


LujieDuan (approved):
LGTM!

PR #445, opened 2025-10-28
Update tools directory

braydonk:
To appease our great lord Dependabot this PR updates the dependencies in the tools directory. I ran into issues updating the deps because of golangci-lint, but those docs actually [recommend not to use the tools
pattern](https://golangci-lint.run/docs/welcome/install/#install-from-sources) so I also move the golangci-lint installation to match their recommendations.


braydonk:
The updates were done in other PRs.

PR #468, opened 2025-12-04
[make] Update `update-components` target to keep `opentelemetry-operations-collector` version in sync.

franciscovalentecastro:
The `transformprocessor` and `filterprocessor` use the "root repo module" [1]  for dependencies on the `/third_party` folder. Keeping the version in sync with the repo will enable this component modules to also work with `go get` when referenced from another module.

Some details :
- We use a replace so this would work locally when developing https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/blob/v0.138.0/components/otelopscol/processor/transformprocessor/go.mod
- Context : https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/350

[1] `github.com/GoogleCloudPlatform/opentelemetry-operations-collector`

### DIFF
--- a/make/otel_component.mk
+++ b/make/otel_component.mk
@@ -37,17 +37,12 @@ core-components:
 		otel_component_versions -otel_version $(OTEL_VERSION) | \
 		$(GO_GET_ALL)
 
-.PHONY: contrib-components
-contrib-components:
-	$(LIST_DIRECT_MODULES) | \
-		$(INCLUDE_CONTRIB_COMPONENTS) | \
-		$(GO_GET_ALL)@$(OTEL_CONTRIB_VERSION)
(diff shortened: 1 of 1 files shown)


make/otel_component.mk
```diff
@@ -41,3 +42,9 @@ contrib-components:
 	$(LIST_DIRECT_MODULES) | \
 		$(INCLUDE_CONTRIB_COMPONENTS) | \
 		$(GO_GET_ALL)@$(OTEL_CONTRIB_VERSION)
+
+.PHONY: operations-collector-components
+contrib-components:
```
braydonk:
```suggestion
operations-collector-components:
```

PR #487, opened 2026-01-26
distrogen: Fix `--compare` feature

braydonk:
I'm sure this worked at one point, I don't know when it broke. The compare code had a bit where it tries to get the content of the existing distribution and the temporary generated one, and at some point I made it so it just fetches the existing directory twice meaning it always succeeded. This has been changed to work as intended.

As of submission, I expect the PR to fail CI and start passing after my next PR that regenerates GBOC distro folder is merged and rebased here.

### DIFF
--- a/distros.yaml
+++ b/distros.yaml
@@ -33,7 +33,6 @@ targets:
         test_distros:
           representative:
           - debian-cloud:debian-11
-          - ml-images:common-gpu-debian-11-py310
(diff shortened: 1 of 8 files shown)


braydonk:
This has a failing check because a PR that would have been blocked by the comparison working was submitted. I am merging this one and will open a followup PR that fixes the issue.

PR #525, opened 2026-04-02
[Testing] Replace depracated DLVM representitive image

LujieDuan:
[b/498959379](http://b/498959379)


integration_test/gce-testing-internal/gce/gce_testing.go
```diff
@@ -1144,7 +1144,7 @@ func addFrameworkMetadata(imageSpec string, inputMetadata map[string]string) (ma
 		}
 		// TODO(b/380470389): we actually *can't* do RunRemotely() on DLVM images due to a bug.
 		// The workaround for the bug is to deploy a fix in-VM via startup scripts.
-		if strings.Contains(imageSpec, "common-gpu-debian-11-py310") {
+		if strings.Contains(imageSpec, "common-cu124-debian11-python310") {
 			metadataCopy["startup-script"] = fmt.Sprintf(`
```
jefferbrecht: We should see if this hack is still needed on the new image

LujieDuan: Tested and looks like we don't need this anymore! Removed the section

jefferbrecht (approved):

PR #529, opened 2026-04-08
feat: add otlphttp exporter and googleclientauth to standard and COS configurations for trace exporting

ridwanmsharif:
b/496920766#comment8


google-built-opentelemetry-collector/docs/examples/configuration/config-standard.yaml
```diff
@@ -87,6 +93,13 @@ exporters:
   # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlemanagedprometheusexporter
   googlemanagedprometheus:
 
+  # The otlphttp exporter is used to send traces to Google Cloud Trace using OTLP http/proto
+  otlphttp:
```
XuechunHou: We should prefer otlpgrpc exporter over the http one, as thats what we use in the Ops Agent and recommend to GKE as well: https://screenshot.googleplex.com/7z7FTtqjCBNQ3BF

LujieDuan (approved):
LGTM!

google-built-opentelemetry-collector/docs/examples/configuration/config-standard.yaml
XuechunHou: Here is a sample otlp grpc exporter config: https://paste.googleplex.com/5467437794459648#l=6

PR #543, opened 2026-04-10
distrogen: Permanent OCB Directory

braydonk:
This PR adds `permanent_ocb_directory` and `vendor_dependencies` features to distrogen. When enabled, it alters the build mode to persist a permanent copy of the generated OCB directory and optionally vendor dependencies inside of it. Alterations were made to the project generation to include update the permanent OCB directory when updating the OTel Version. More details can be found in `docs/distrogen/distribution.md`.


cmd/distrogen/templates/distribution/Makefile.go.tmpl
```diff
@@ -3,11 +3,12 @@
 DISTRIBUTION_NAME = {{ .Name }}
 
 # Variables filled in from distro specification
-VERSION ?= {{ .Version }}
-OTEL_VERSION ?= {{ .OpenTelemetryVersion }}
+VERSION = {{ .Version }}
```
braydonk, line 6: These two lines actually were an intentional change even though they don't look like it. Them being `?=` is an artifact of old experimental versions of the tool. The way it works now, it never makes sense to override these values when calling any make target.

braydonk:
Damn, we need an OCB feature. I'll submit a PR upstream and see if they'll take it.

braydonk:
Opened https://github.com/open-telemetry/opentelemetry-collector/issues/15097 upstream with the feature that we need to make this work.

google-built-opentelemetry-collector/spec.yaml
```diff
@@ -120,6 +120,8 @@ replaces:
 feature_gates:
     - exporter.googlemanagedprometheus.intToDouble
     - exporter.googlecloud.CustomMonitoredResources
+permanent_ocb_directory: true
 collector_cgo: true
 component_module_base: ""
 distrogen_version: ""
+ocboutputdir: generated_collector
```
avilevy18: Nit: `ocboutputdir` should be `ocb_output_dir` for readability.

google-built-opentelemetry-collector/spec.yaml
```diff
@@ -120,6 +120,8 @@ replaces:
 feature_gates:
     - exporter.googlemanagedprometheus.intToDouble
     - exporter.googlecloud.CustomMonitoredResources
+permanent_ocb_directory: true
```
avilevy18, line 123: we should be consistent in our naming, of using `dir` or `directory`

google-built-opentelemetry-collector/spec.yaml
avilevy18: `OCBOutputDir` doesn't have yaml binding. Should this be removed?

braydonk: The intention was for it to be completely ignored by yaml; I forgot to put `yaml:"-"` for it. Done in [2f78b61](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/543/commits/2f78b61e81826c7bedd3738e8dccbedc92fb00cc)

avilevy18 (approved):

PR #544, opened 2026-04-10
Fix missed generate

braydonk:
A previous PR missed the generation step for the GBOC distribution. This was missed while the `--compare` feature of distrogen was broken, meaning our comparison presubmit always passed.


braydonk:
Nevermind I realize what happened now; the changes were done backwards, directly on the generated files and not on the templates. Let's try that again.

PR #550, opened 2026-04-21
Wait for background guestregister service

jefferbrecht:
There is an existing service on SLES VMs that runs `registercloudguest`, so wait for it first before running it ourselves as a fallback to avoid clashes. http://b/503363636


LujieDuan (approved):
LGTM!

PR #553, opened 2026-04-30
Update GBOC to v0.151.0

LujieDuan:
- CERT_CONTAINER and BUILD_CONTAINER are up to date
- [b/508218445](http://go/508218445)


components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter/internal/metadata/generated_status.go
```diff
@@ -1,5 +1,7 @@
 // Code generated by mdatagen. DO NOT EDIT.
 
+// Package metadata contains the autogenerated telemetry and
```
braydonk, line 3: Was this needed for a lint failure?

braydonk (approved):

components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter/internal/metadata/generated_status.go
braydonk: Looks like mdatagen actually generates this as part of the template now. Not called out in release notes but I guess it doesn't really matter?

PR #559, opened 2026-05-25
feat(distrogen): Update distrogen if it detects different version in spec.yaml

avilevy18:
This pull request updates the generated project's Makefile template so that the `$(DISTROGEN_BIN)` target depends on the spec file (`$(SPEC_FILE)`). 

Previously, the `$(DISTROGEN_BIN)` target had no file dependencies. Once `.tools/distrogen` was created on disk, `make` would always consider the target up-to-date and completely skip its recipe, even if the developer updated `distrogen_version` in `spec.yaml`.


cmd/distrogen/templates/project/Makefile.go.tmpl
```diff
@@ -16,9 +16,13 @@ DISTROGEN_VERSION ?= $(shell $(DISTROGEN_QUERY) distrogen_version)
 MDATAGEN_BIN ?= $(TOOLS_DIR)/mdatagen
 DISTRO_DIR = $(PWD)/{{ .Spec.Name }}
 
-$(DISTROGEN_BIN):
+$(DISTROGEN_BIN): $(SPEC_FILE)
```
braydonk, line 19: Doesn't this mean distrogen will update every single time the spec does though?

avilevy18: From my understanding it means it will attempt to update every single time there is a change in spec.yaml. 

If the current version is the same as the `distrogen_version` then it will just skip it.

avilevy18 (commented):
Closing since https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/561 fixes the same thing in a better way.

PR #560, opened 2026-05-26
Fix project makefile for vendored deps mode

braydonk:
This fixes up the flow in the project makefile so that the OCB dir update happens as part of the generation step instead of the otel updating step (but it will still happen as part of that since it's within the `gen` flow).


avilevy18 (approved):
LGTM

PR #561, opened 2026-05-28
distrogen: detect necessary tool changes

braydonk:
When I refactored the generation to not clear the original generated directory, one edge case I didn't consider is that the `.tools` directory that was there won't be cleared. This will cause the Makefile to sometimes not decide to install a new version of the tool when it actually needs to (such as changing Go version or the new OTel version requiring reinstall of `ocb`).

This PR makes a change to the `DistributionGenerator` to detect a spec change that will require tool reinstallation, and will specifically clean the existing `.tools` directory if necessary.

I evaluated detecting and deleting specific tool reinstallation and deletion, but it is far more complex for very minimal gain vs just wiping the `.tools` directory when any tool needs to change.


avilevy18 (commented):
Can we also do this for the project level `.tools` on the project level as well? This should take care of: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/559 then.

braydonk:
> Can we also do this for the project level .tools on the project level as well?

I had to add functionality for spec caching in project generator the same as distribution generator. I've done that and got the tool change detection working basically the same way. I tested it manually, and the `$(DISTROGEN_BIN)` target correctly detected that it needed to be redownloaded when I tested version bumping manually.

avilevy18 (approved):
LGTM

PR #567, opened 2026-06-10
docs(distrogen): document custom component generation

avilevy18:
This PR updates the `distrogen` documentation to make the custom components usage clearer and provides a better example of how the command works. It also fixes a template rendering bug found related to the Go major version so that the `component` command executes successfully.

### DIFF
--- a/cmd/distrogen/distribution.go
+++ b/cmd/distrogen/distribution.go
@@ -86,8 +86,8 @@ type DistributionSpec struct {
 	OCBOutputDir string `yaml:"-"`
 }
 
-// RenderGoMajorVersion will parse the GoVersion in the spec and return a version without a patch
-func (s *DistributionSpec) RenderGoMajorVersion() string {
+// GoMajorVersion will parse the GoVersion in the spec and return a version without a patch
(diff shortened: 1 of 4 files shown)


cmd/distrogen/distribution.go
```diff
@@ -86,8 +86,8 @@ type DistributionSpec struct {
 	OCBOutputDir string `yaml:"-"`
 }
 
-// RenderGoMajorVersion will parse the GoVersion in the spec and return a version without a patch
-func (s *DistributionSpec) RenderGoMajorVersion() string {
+// GoMajorVersion will parse the GoVersion in the spec and return a version without a patch
+func (s *DistributionSpec) GoMajorVersion() string {
```
braydonk: Why?

braydonk (approved):

cmd/distrogen/distribution.go
avilevy18: Spoke offline, we decided to go back to `RenderGoMajorVersion`

braydonk: I was getting tricked up by a very strange diff. We got it sorted offline

PR #570, opened 2026-06-12
Recognize AlmaLinux as RHEL

XuechunHou:
Updates isRPM and isRPMBased to correctly recognize AlmaLinux as an RPM-based OS instead of falling back to Debian/apt-get in tests.


integration_test/smoke_test/smoke_test.go
```diff
@@ -166,6 +166,7 @@ func restartCommandForPlatform(platform string) string {
 func isRPMBased(imageSpec string) bool {
 	return strings.HasPrefix(imageSpec, "centos-cloud") ||
 		strings.HasPrefix(imageSpec, "rhel-") ||
+		strings.Contains(imageSpec, "almalinux-") ||
```
LujieDuan, line 169: Nit: probably good enough with HasPrefix() here to match other lines

LujieDuan (approved):

PR #572, opened 2026-06-12
feat(distrogen): add distroless Dockerfile for distrogen build

avilevy18:
Create a distorgen Dockerfile

### DIFF
--- a/cmd/distrogen/Dockerfile
+++ b/cmd/distrogen/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
 
-COPY . .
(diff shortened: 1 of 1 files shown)


cmd/distrogen/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM golang:1.26 AS builder
+
+WORKDIR /app
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
```
LujieDuan:
```suggestion
COPY ./cmd ./cmd
```
Nit: would this work? Might save a bit time by not copying the entire repo

LujieDuan (approved):

PR #585, opened 2026-06-19
distrogen: make updating the Go module in generated collector purposeful

braydonk:
The default entrypoints for regenerating the distribution will no longer always update the `go.mod` file when using `permanent_ocb_directory`. This means that by default, the go mod file that starts there will remain there when regenerating. You now need to ask specifically to update the `go.mod` file by either setting the `MODULE_UPDATES` make variable in the environment or using the `update-ocb-directory` target. The project makefile has been updated so that `update-otel-components` will trigger a Go module update in the generated collector.


cmd/distrogen/templates/distribution/Makefile.go.tmpl
```diff
@@ -121,18 +121,27 @@ endif
...
+{{- end }}
+
 # If the COLLECTOR_BUILD_DIR is set to the permanent OCB directory
 # value, then this will ensure that the permanent OCB directory is
 # not updated. If COLLECTOR_BUILD_DIR is set to the same value as
 # COLLECTOR_BUILD_TEMP_DIR, then the OCB directory will be regenerated
 # since the collector build target will always clean up the _build directory.
 $(COLLECTOR_BUILD_DIR):
-	$(MAKE) ocb-generate
+	MODULE_UPDATE=true $(MAKE) ocb-generate
```
avilevy18: why is this always suppose to be true?

braydonk: It's a bit clearer now with the comment and commit message, but to clarify: the intention of this target is that if the directory doesn't exist, we have to call OCB without `--skip-get-modules` to make sure the initial Go mod gets generated.

cmd/distrogen/templates/project/Makefile.go.tmpl
```diff
@@ -43,12 +44,16 @@ COMPONENT_DIR = components
...
-	export OTEL_CONTRIB_VERSION = v$(shell $(DISTROGEN_QUERY) opentelemetry_contrib_version) && \
+	export OTEL_VERSION=v$(shell $(DISTROGEN_QUERY) opentelemetry_version) && \
+	export OTEL_CONTRIB_VERSION=v$(shell $(DISTROGEN_QUERY) opentelemetry_contrib_version) && \
 	export PATH="$(TOOLS_DIR):${PATH}" && \
 	cd components && \
 	$(MAKE) update-components
+{{- if .Spec.PermanentOCBDirectory }}
+	$(MAKE) gen-update
+{{- else }}
 	$(MAKE) gen
```
avilevy18, line 55:
```suggestion
	$(MAKE) update-components && \
{{- if .Spec.PermanentOCBDirectory }}
	$(MAKE) gen-update && \
{{- else }}
	$(MAKE) gen
```

avilevy18 (approved):

cmd/distrogen/templates/project/Makefile.go.tmpl
braydonk: The && isn't necessary; that's only when you need to keep all commands in the same shell context within make (i.e. when you've changed directories). Since these can all run from the same wd they can just be separate lines

avilevy18: Gotcha, LGTM

PR #588, opened 2026-06-25
feat: Support multiple build tags in distrogen and disable unused Prometheus discoveries

rafaelwestphal:
### Description
This PR resolves a security compliance finding regarding the EOL `github.com/aws/aws-sdk-go` (v1) dependency in the `otelopscol` binary.

The EOL dependency was pulled in transitively via the Prometheus receiver's Outscale service discovery plugin:
`prometheusreceiver` -> `prometheus/prometheus` -> `outscale/osc-sdk-go/v2` -> `aws-sdk-go` (v1).

Since the Google Cloud Ops Agent does not require Outscale service discovery, we can safely disable it at build time using Go build tags, avoiding the need for any code forks.


cmd/distrogen/testdata/generator/distribution/basic/golden/spec.yaml
```diff
@@ -10,7 +10,7 @@ opentelemetry_contrib_version: 0.121.0
 opentelemetry_stable_version: 1.27.0
 go_version: 1.24.0
 binary_name: otelcol-basic
-build_tags: ""
+build_tags: []
```
braydonk: I actually kinda wanted to just provide these the same way you would in the actual command, like a comma separated string.

braydonk (approved):

PR #589, opened 2026-06-25
distrogen: distrogen auto-updating

braydonk:
After a project is bootstrapped, the changes in this PR should make it so that users never again need to manually install distrogen. Bootstrapping based on the `distrogen_version` field in the spec should be built-in.


avilevy18 (approved):

cmd/distrogen/testdata/generator/project/basic/golden/.distrogen/VERSION
```diff
@@ -1 +1 @@
-v0.130.1-0.20250724183747-261516e1bff1
+(devel)
```
avilevy18, line 1: sorry can you explain this?

braydonk: This file will now reflect what it's supposed to, which is the version of distrogen used to generate the project. It reads now from the built binary, and the test binary doesn't end up coming with any `buildvcs` version info, but any `go install`ed distrogen will come with the tag that was used

avilevy18: Thanks!!!

avilevy18:
We should update, `docs/distrogen/command.md ` with the new version command.

avilevy18:
Can we make a new markdown file for updating a project? Or updating distrogen?

PR #593, opened 2026-07-06
googleservicecontrolexporter: fix flaky CI test

braydonk:
[This recently flaky test](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/actions/runs/28467756544/job/84371892093?pr=587) is caused by `google.NewDefaultCredentials()` which is called during exporter creation. In CI environments without local GCP credentials, this function attempts to check for the GCE metadata server. This check establishes a persistent HTTP connection cached in `http.DefaultTransport`'s connection pool. These idle connection goroutines remained alive at the end of the test run, causing goleak to fail.

This PR mocks the credential generation in unit tests:

* Wrapped GCP credentials initialization behind a package-level variable getCredentials in `factory.go`
* Overrode getCredentials in `factory_test.go`'s init() function with a mock bundle that uses insecure.NewCredentials() and does not hit the network.


LujieDuan (approved):
LGTM!

PR #602, opened 2026-07-15
Add presubmit validation check for sample configuration files

rafaelwestphal:


.github/workflows/presubmit.yaml
LujieDuan, line 29: I believe this step should invoke the validation step already by calling `make presubmit` - so we don't need a new job config here

LujieDuan: I misread - `validate-samples` is added to `precommit`, not `presubmit`.

LujieDuan (approved):

PR #607, opened 2026-07-20
Retry VM starts on GCE zone capacity stockouts and availability errors

rafaelwestphal:
Matching the behavior on CreateInstance.
This will prevent some Arm tests from failing because of ZONE_RESOURCE_POOL_EXHAUSTED when restarting the VM


LujieDuan (approved):
LGTM

PR #609, opened 2026-07-23
test(gce): Deflake SSH remote execution with fast timeouts, transport retries, and transport error handling

rafaelwestphal:
Deflake GCE integration test SSH executions by introducing fast connect timeouts, OpenSSH connection retries, and Go harness backoff retries for transient SSH errors.

### Problem
During parallel test runs, temporary network congestion can drop SSH connections to GCE VMs.

Previously:
1. `sshOptions` used `-oConnectTimeout=120` with zero retries. When a TCP SYN dropped, `ssh` hung for 2 full minutes before exiting with code 255.
2. `RunRemotelyStdin` ran `ssh` exactly once with no retry loop, causing transient connection drops to fail the entire test immediately with `exit status 255`.
3. Functions like `verifyGcloudInstallation` misinterpreted SSH transport connection failures (`exit status 255`) as missing OS binaries, leading to misleading error messages (`"this test does not know how to install gcloud on image spec: debian"`).

### Solution
1. **Faster Connect Timeout & Built-in SSH Retries:**
   * Updated `sshOptions` to `-oConnectTimeout=15` and `-oConnectionAttempts=3`. Dropped TCP SYN packets now fail fast in 15 seconds inside OpenSSH rather than hanging for 120 seconds.
2. **Go Harness Backoff Retry for Transport Errors:**
   * Added `isSSHTransportError()` helper to detect OpenSSH client transport failures (`exit status 255`, connection timeouts, refused connections).
   * Wrapped `RunRemotelyStdin` in a `backoff.Retry` loop (3 attempts with 3s backoff). Non-SSH transport errors (such as remote command non-zero exit codes) are immediately returned via `backoff.Permanent(runErr)`.
3. **Improved Diagnostics:**
   * Updated `InstallGcloudIfNeeded` to check `isSSHTransportError`, preventing SSH transport drops from triggering unnecessary fallback installation steps.

### Testing
- Added `TestIsSSHTransportError` in `gce_testing_test.go` covering `nil` errors, generic command failures (exit code 1), SSH connection timeouts, connection refused, and exit status 255.
- Verified local package compilation and unit test execution (`PASS: TestIsSSHTransportError`).
- Verified proper error propagation for non-zero remote command exit codes (e.g. `TestInvalidConfig`).

Opened a PR on OpsAgent that used this and had all tests pass (mostly).
https://github.com/GoogleCloudPlatform/ops-agent/pull/2391/


integration_test/gce-testing-internal/gce/gce_testing.go
```diff
@@ -882,9 +882,32 @@ func RunRemotely(ctx context.Context, logger *log.Logger, vm *VM, command string
 	return RunRemotelyStdin(ctx, logger, vm, nil, command)
 }
 
+// isSSHTransportError returns true if the error was caused by an SSH connection/transport failure
+// (e.g. TCP SYN timeout, connection dropped/refused) rather than a command failing on the remote host.
+func isSSHTransportError(out CommandOutput, err error) bool {
```
LujieDuan: `out` is not used in the function - can be removed from the input.

integration_test/gce-testing-internal/gce/gce_testing.go
```diff
@@ -838,9 +838,9 @@ func RunGcloud(ctx context.Context, logger *log.Logger, stdin string, args []str
 
 var (
 	sshOptions = []string{
-		// In some situations, ssh will hang when connecting to a new VM unless
-		// it has an explicit connection timeout set.
-		"-oConnectTimeout=120",
+		// Fail fast on dropped SSH TCP SYN packets instead of hanging for 120s.
+		"-oConnectTimeout=15",
+		"-oConnectionAttempts=3",
```
LujieDuan, line 843: Question: Would `-oConnectionAttempts=3` be a duplication with the go retry? 
e.g., if the SSH connection keeps dropping, aren't we doing 15s x 3 attempts x (1 initial call + 3 go retries) = 180s in total?

rafaelwestphal: Yeah, in practice:
Scenario	                                                        /Before	                                       / After
Transient TCP SYN Drop	/120s hang then Test Failure	/Resolves in ~15–18s (Success on 2nd attempt)
VM Completely Unreachable / Dead	        /Fails after 120s	                        /Fails after 189s

LujieDuan: I see - it is the first scenario that we are focusing on here for fast fail. LGTM!

LujieDuan (approved):

PR #614, opened 2026-07-31
Copy jmxreceiver at 0.156.0; adding otel license headers where missing

LujieDuan:
Add jmxreceiver vendor copy

[b/483465937](http://b/483465937)
Copied from [open-telemetry/opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.156.0/receiver/jmxreceiver)
Version: v0.156.0
Commit: 41e24cd516dd69a5b4277465cdb2ff4ef0676f49
Tested in: https://github.com/GoogleCloudPlatform/ops-agent/pull/2395


franciscovalentecastro:
Could we possibly just point the `jmxreceiver` go component to upstream `v0.155.0` (or before it was deleted) to avoid having to copy ?

LujieDuan:
> Could we possibly just point the `jmxreceiver` go component to upstream `v0.155.0` (or before it was deleted) to avoid having to copy ?

I believe it would work for recent versions, but not when the dependencies have breaking changes; because the jmxreceiver code has dependencies on the collector modules, it can't have an earlier version of those modules while the rest of the otelopscol requires a later version.

rafaelwestphal (approved):

PR #619, opened 2026-08-11
Wait for snap seeding before verifying gcloud in GCE integration tests

rafaelwestphal:
On Ubuntu 24.04 and ML images, gcloud is packaged as a snap. During first boot of a fresh VM, snapd.seeded.service runs asynchronously to unpack and mount google-cloud-cli. This caused tests calling UploadContent immediately upon SSH connection to fail flakily with 'sudo: gcloud: command not found'.

This adds a check to wait for snap seeding ('sudo snap wait system seed.loaded') if snap is installed, and adds backoff retries to verifyGcloudInstallation.

b/545156163


LujieDuan (approved):
LGTM!

PR #631, opened 2026-09-02
Add canonical telemetry policy and xDS discovery definitions

dashpole:
## Overview
This PR introduces the canonical Protobuf definitions, repository structure, code generation, and CI verification for **Telemetry Policies** and their **xDS discovery transport** in `opentelemetry-operations-collector`.

It introduces `LogFilterPolicy` as an initial schema stub to establish the package namespace, directory structure, and xDS extension points, with specific filter fields and matchers deferred to a follow-up PR.

---

### 1. Canonical Protobuf Definitions (`proto/`)
- **Tier 1: Protocol-Agnostic Telemetry Policies (`proto/policy/v1alpha1/`)**:
  - `log_filter_policy.proto`: Defines `LogFilterPolicy` under package `google.telemetry.policy.v1alpha1`. This schema represents declarative intent for filtering log records and contains zero transport or networking dependencies.
- **Tier 2: xDS Wire Resource (`proto/xds/v1alpha1/`)**:
  - `telemetry_collector.proto`: Defines the top-level xDS configuration resource `TelemetryCollector` (`type.googleapis.com/google.telemetry.xds.v1alpha1.TelemetryCollector`).
  - Encapsulates policies using standard Envoy `repeated envoy.config.core.v3.TypedExtensionConfig policies = 1`, where each policy is wrapped in a `google.protobuf.Any`.
  - Documented to align encapsulated policy types with [OTEP 4738 telemetry policy definitions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4738-telemetry-policy.md) and [OpAMP custom messages](https://opentelemetry.io/docs/specs/opamp/#custom-messages).
  - Omitted resource-level `name` field in favor of client node/fleet identification via standard `DiscoveryRequest`.
- **Tier 3: xDS Discovery Contract (`proto/xds/v1alpha1/`)**:
  - `telemetry_discovery_service.proto`: Defines `TelemetryDiscoveryService` for `TelemetryCollector` resources, providing point-to-point discovery RPCs (`StreamTelemetryCollectors`, `DeltaTelemetryCollectors`, and `FetchTelemetryCollectors`) using standard Envoy `DiscoveryRequest` and `DiscoveryResponse` messages.
  - Documents a recommendation for clients to prefer Aggregated Discovery Service (ADS) for resource multiplexing, and notes compatibility with Cloud Service Mesh.

### 2. Code Generation & Repository Layout (`gen/go/`)
[...]

PR #632, opened 2026-09-02
Add LogFilterPolicy definition and generated Go bindings

dashpole:
## Description

This PR defines the canonical schema and generated Go bindings for **`LogFilterPolicy`** under `google.telemetry.policy.v1alpha1`, following the merged xDS transport foundation (#631).

`LogFilterPolicy` provides a declarative, transport-agnostic specification for retaining or dropping OpenTelemetry log records.

### Schema Summary
* **`LogFilterPolicy`**: Top-level atomic policy containing an identifier (`id`), an explicit optional action (`action`), and a list of ANDed condition matchers (`matches`). At least one matcher is required to prevent accidental drop-all policies.
* **`Action`**: Strongly typed enum (`ACTION_UNSPECIFIED = 0`, `ACTION_KEEP = 1`, `ACTION_DROP = 2`), marked `optional` per AIP-203 to distinguish omitted action from `ACTION_UNSPECIFIED`. An unspecified action is invalid; fail-open evaluation skips invalid policies at runtime with errors reported.
* **`LogMatcher`**: Evaluates a `LogFieldSelector` against a `oneof predicate` (`exists`, `exact`, or `regex` using RE2 syntax), with a boolean `negate` flag on tag 2.
* **`LogFieldSelector`**: Strongly typed target selecting standard `record_field`s (`BODY`, `SEVERITY_TEXT`, `SEVERITY_NUMBER`, `TRACE_ID`, `SPAN_ID`), flat attribute keys (`log_attribute`, `resource_attribute`, `scope_attribute`), scope metadata via `ScopeField` enum (`scope_field`), or resource metadata via `ResourceField` enum (`resource_field`).

---

## Comparison with OTEP 4738

This schema aligns with the core principles of [OTEP 4738](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4738-telemetry-policy.md) (declarative, transport-agnostic, fail-open), but introduces intentional structural refinements:

* **Atomic Filtering vs. Multi-Action Envelope**: OTEP 4738 combines filtering, rate-limiting, percentage sampling, and mutations (`redact`/`transform`) into a single object (`"keep": "none"`, `"keep": "50%"`). This PR establishes `LogFilterPolicy` as an atomic boolean filtering policy (`KEEP` vs. `DROP`). Mutations and sampling will be modeled as separate policy types.
* **Typed `Action` Enum vs. Polymorphic String**: OTEP 4738 represents policy behavior via a polymorphic string field (`"keep": "none"` / `"all"` / `"N%"` / `"N/s"`). This PR uses an `optional Action` enum, eliminating string parsing, casing bugs, and ambiguous states in favor of compile-time type safety.
* **`google.protobuf.Empty` for `exists` vs. `bool exists`**: OTEP 4738 examples specify `"exists": true`. In Protobuf, putting `bool exists` inside a `oneof predicate` allows an ambiguous `exists: false` state alongside `negate: true`. Using `google.protobuf.Empty` eliminates this ambiguity: setting `exists` tests for presence, while setting `exists` with `negate: true` tests for absence.
* **Structured Scope and Resource Enums vs. Open Fields**: OTEP 4738 groups schema URLs into `LogField` without dedicated scope selectors. This PR introduces `ScopeField` (`NAME`, `VERSION`, `SCHEMA_URL`) and `ResourceField` (`SCHEMA_URL`) enums, providing strongly typed, extensible selectors while avoiding boolean presence flags in `oneof target`.
* **Flat Attribute Keys vs. Key Arrays**: OTEP 4738 examples use arrays for attribute paths (e.g. `["ccn"]`). This PR uses flat strings (`string log_attribute = 2;`, etc.) to optimize for the vast majority of OTel attributes without heap allocations. Nested path traversal is deferred and can be added later as a separate variant in `oneof target`.
* **String Coercion Specifications**: To prevent evaluator discrepancies across collector implementations, this PR explicitly documents string coercion for non-string fields (`SEVERITY_NUMBER` as base-10 decimal `"17"`, and `TRACE_ID`/`SPAN_ID` as lowercase hex strings).
* **Omission of `name` / `description`**: OTEP 4738 includes human-readable `"name"` and `"description"` fields on policies. In our architecture, policies are wrapped inside `envoy.config.core.v3.TypedExtensionConfig`, where `name` is already the extension identifier. Human-readable descriptions are maintained in management plane APIs (`PolicySet`), keeping wire policy protos minimal.
* **At Least One Matcher Required**: OTEP 4738 specifies that implementations must reject policies with empty matchers. This PR documents this constraint explicitly in the proto to prevent accidental drop-all policies when `ACTION_DROP` is selected.

---

## Alternatives Considered

1. **Open String / Boolean Action (Rejected)**
   * *Considered*: `string action = 2;` or `bool drop = 2;`.
   * *Why Rejected*: Strings invite typo and casing bugs and lack compile-time validation (AIP-126). Booleans lack an explicit `UNSPECIFIED` state and prevent future extension. An enum provides type safety, compile-time constants, and clear invalid/unspecified handling.
2. **`bool exists` vs. `google.protobuf.Empty exists` in `oneof predicate` (Rejected)**
   * *Considered*: `bool exists = 10;`.
   * *Why Rejected*: Allows setting `exists: false`, which creates an ambiguous wire state alongside `bool negate`. Using `google.protobuf.Empty` ensures existence is expressed solely by presence in `oneof`, and absence is expressed by pairing with `negate: true`.
3. **Boolean Flags vs. Enums for Scope/Resource Targets (Rejected)**
   * *Considered*: `bool scope_name = 5;`, `bool scope_version = 6;`, etc.
   * *Why Rejected*: Boolean flags in a `oneof` allow an ambiguous `false` state and triggered AIP-122 linter warnings. Grouping them into `ScopeField` and `ResourceField` enums is cleaner, strongly typed, and extensible.
4. **Unstructured Expression Strings / OTTL / CEL (Rejected)**
   * *Considered*: Defining filter criteria as expression strings (e.g. `attributes["http.route"] == "/healthz"`).
   * *Why Rejected*: Defeats Protobuf’s strongly typed structure, introduces escaping and quoting edge cases, couples the schema to specific expression grammar runtimes, and forces all components to embed expression parsers.
5. **Operator Enum + Value Pair (Rejected)**
   * *Considered*: An enum `Operator { EQUALS = 0; REGEX = 1; EXISTS = 2; }` paired with a `string value = 3;`.
   * *Why Rejected*: Leaves `value` ambiguous and unused for `EXISTS`, and allows invalid state combinations. A `oneof predicate` makes invalid states unrepresentable on the wire while `bool negate` provides clean logical inversion.
6. **Recursive Boolean Expression Trees (Rejected)**
   * *Considered*: Full boolean trees (`all_of`, `any_of`, `none_of`).
   * *Why Rejected*: Turns declarative configuration into a programming language, increasing evaluation latency and complexity. A flat list of ANDed matchers satisfies standard filtering needs; OR semantics are achieved by defining multiple policies in a policy set.

---

## Deferred Features

1. **Nested Attribute Paths**: 99% of OTel attributes use flat string keys (`"http.route"`). A `repeated string path` / `AttributePath` variant can be added to `LogFieldSelector` later without breaking wire compatibility.
2. **Sampling Actions (`ACTION_SAMPLE`)**: Boolean filtering is kept distinct from probabilistic sampling and rate limiting. Sampling can be added as a dedicated action or independent policy type (`LogSamplingPolicy`).
3. **Numeric Range Predicates (`gt`, `gte`, `lt`, `lte`)**: Log filtering is overwhelmingly string- and attribute-based. Numeric comparisons can be introduced as dedicated scalar predicate variants if needed for latency or numerical status thresholds.
4. **Syntactic Sugar Matchers (`starts_with`, `ends_with`, `contains`)**: Prefix, suffix, substring, and case-insensitive matching are already supported via RE2 regexes and exact matching. Dedicated predicate variants can be added later if profiling demonstrates measurable performance gains.
5. **Cross-Policy Priority / Conflict Resolution**: Multi-policy resolution is managed at the policy set / fleet envelope level rather than inside atomic policy definitions.

### DIFF
--- a/proto/policy/v1alpha1/log_filter_policy.proto
+++ b/proto/policy/v1alpha1/log_filter_policy.proto
@@ -26,7 +26,7 @@ option go_package = "github.com/GoogleCloudPlatform/opentelemetry-operations-col
 // Policies are declarative, atomic, and transport-agnostic.
 message LogFilterPolicy {
   // Unique identifier for the policy within its configuration scope
-  // (e.g. "drop-noisy-healthchecks", "retain-error-logs").
+  // (e.g. a UUID or namespaced URI like "io.opentelemetry.operator/{namespace}/policies/retain-error-logs").
   string id = 1;
 
   // The filtering action to take when all `matches` conditions evaluate to true.
--- a/proto/policy/v1alpha1/log_filter_policy.proto
+++ b/proto/policy/v1alpha1/log_filter_policy.proto
@@ -59,7 +59,7 @@ message LogMatcher {
   }
 
   // If true, inverts the result of the predicate (logical NOT).
-  bool negate = 20;
+  bool negate = 2;
 }
 
 // LogFieldSelector designates a first-class OTel log field, attribute, or metadata.
--- a/proto/policy/v1alpha1/log_filter_policy.proto
+++ b/proto/policy/v1alpha1/log_filter_policy.proto
@@ -18,11 +18,108 @@ package google.telemetry.policy.v1alpha1;
 
 option go_package = "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/gen/go/policy/v1alpha1;policyv1alpha1";
 
-// LogFilterPolicy defines an intent-based rule for filtering log records.
+// LogFilterPolicy defines an intent-based rule for retaining or dropping
+// log records based on structured field and attribute matches.
 //
-// Policies are declarative and transport-agnostic.
-//
-// Note: Detailed filter conditions, matchers, and action semantics are deferred
-// to a subsequent design. This message is currently stubbed to establish the
-// package hierarchy, repository organization, and xDS transport plumbing.
-message LogFilterPolicy {}
+// Policies are declarative, atomic, and transport-agnostic.
+message LogFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. "drop-noisy-healthchecks", "retain-error-logs").
+  string id = 1;
+
+  // The filtering action to take when all `matches` conditions evaluate to true.
+  Action action = 2;
+
+  // Conditions evaluated against incoming log records.
+  // All matchers in this list are ANDed together: all conditions must evaluate
+  // to true for the policy action to take effect. If empty, the policy matches
+  // all log records.
+  repeated LogMatcher matches = 3;
+}
+
+// LogMatcher defines a single predicate evaluated against a specific log field
+// or attribute.
+message LogMatcher {
+  // The field, attribute, or metadata to inspect. Exactly one must be set.
+  LogFieldSelector target = 1;
+
+  // The comparison predicate to evaluate against the target value.
+  // Exactly one predicate must be set.
+  oneof predicate {
+    // Matches if the selected field or attribute exists.
+    bool exists = 10;
+
+    // Exact string equality match.
+    string exact = 11;
+
+    // Regular expression match using RE2 syntax.
+    string regex = 12;
+  }
+
+  // If true, inverts the result of the predicate (logical NOT).
+  bool negate = 20;
+}
+
+// LogFieldSelector designates a first-class OTel log field, attribute, or metadata.
+message LogFieldSelector {
+  oneof target {
+    // Standard first-class OpenTelemetry LogRecord fields.
+    LogRecordField record_field = 1;
+
+    // Log record attribute by key (e.g. "http.route", "user_id").
+    string log_attribute = 2;
+
+    // Resource attribute by key (e.g. "service.name", "k8s.pod.name").
+    string resource_attribute = 3;
+
+    // Instrumentation scope attribute by key (e.g. "library.name").
+    string scope_attribute = 4;
+
+    // Matches the Instrumentation Scope name.
+    // (-- api-linter: core::0122::name-suffix=disabled
+    //     aip.dev/not-precedent: Matches OpenTelemetry InstrumentationScope name. --)
+    bool scope_name = 5;
+
+    // Matches the Instrumentation Scope version.
+    bool scope_version = 6;
+
+    // Matches the Resource Schema URL.
+    bool resource_schema_url = 7;
+
+    // Matches the Instrumentation Scope Schema URL.
+    bool scope_schema_url = 8;
+  }
+}
+
+// Action specifies what to do with a log record that matches all conditions.
+enum Action {
+  // Default unspecified action. A policy with an unspecified action is invalid.
+  ACTION_UNSPECIFIED = 0;
+
+  // Keep the matching log record and forward it to downstream pipelines.
+  ACTION_KEEP = 1;
+
+  // Drop the matching log record completely.
+  ACTION_DROP = 2;
+}
+
+// LogRecordField identifies standard first-class fields of an OpenTelemetry LogRecord.
+enum LogRecordField {
+  // Default unspecified log record field.
+  LOG_RECORD_FIELD_UNSPECIFIED = 0;
+
+  // The primary message body of the log record.
+  LOG_RECORD_FIELD_BODY = 1;
+
+  // The string severity of the log (e.g. "INFO", "ERROR", "DEBUG").
+  LOG_RECORD_FIELD_SEVERITY_TEXT = 2;
+
+  // The numerical severity of the log (1-24 per OpenTelemetry specification).
+  LOG_RECORD_FIELD_SEVERITY_NUMBER = 3;
+
+  // The 16-byte trace identifier associated with the log record.
+  LOG_RECORD_FIELD_TRACE_ID = 4;
+
+  // The 8-byte span identifier associated with the log record.
+  LOG_RECORD_FIELD_SPAN_ID = 5;
+}
--- a/proto/policy/v1alpha1/log_filter_policy.proto
+++ b/proto/policy/v1alpha1/log_filter_policy.proto
@@ -109,12 +109,20 @@ enum LogRecordField {
   LOG_RECORD_FIELD_SEVERITY_TEXT = 2;
 
   // The numerical severity of the log (1-24 per OpenTelemetry specification).
+  // When evaluated against string predicates (`exact`, `regex`), this matches
+  // against the canonical base-10 integer string representation (e.g. "17").
   LOG_RECORD_FIELD_SEVERITY_NUMBER = 3;
 
   // The 16-byte trace identifier associated with the log record.
+  // When evaluated against string predicates (`exact`, `regex`), this matches
+  // against the 32-character lowercase hexadecimal string representation
+  // per the OpenTelemetry / W3C Trace Context specification.
   LOG_RECORD_FIELD_TRACE_ID = 4;
 
   // The 8-byte span identifier associated with the log record.
+  // When evaluated against string predicates (`exact`, `regex`), this matches
+  // against the 16-character lowercase hexadecimal string representation
+  // per the OpenTelemetry / W3C Trace Context specification.
   LOG_RECORD_FIELD_SPAN_ID = 5;
 }
 
--- a/proto/policy/v1alpha1/log_filter_policy.proto
+++ b/proto/policy/v1alpha1/log_filter_policy.proto
@@ -34,8 +34,8 @@ message LogFilterPolicy {
 
   // Conditions evaluated against incoming log records.
   // All matchers in this list are ANDed together: all conditions must evaluate
-  // to true for the policy action to take effect. If empty, the policy matches
-  // all log records.
+  // to true for the policy action to take effect.
+  // At least one matcher is required; a policy with no matchers is invalid.
   repeated LogMatcher matches = 3;
 }
 
--- a/proto/policy/v1alpha1/log_filter_policy.proto
+++ b/proto/policy/v1alpha1/log_filter_policy.proto
@@ -77,19 +77,11 @@ message LogFieldSelector {
     // Instrumentation scope attribute by key (e.g. "library.name").
     string scope_attribute = 4;
 
-    // Matches the Instrumentation Scope name.
-    // (-- api-linter: core::0122::name-suffix=disabled
-    //     aip.dev/not-precedent: Matches OpenTelemetry InstrumentationScope name. --)
-    bool scope_name = 5;
+    // Standard first-class OpenTelemetry InstrumentationScope fields.
+    ScopeField scope_field = 5;
 
-    // Matches the Instrumentation Scope version.
-    bool scope_version = 6;
-
-    // Matches the Resource Schema URL.
-    bool resource_schema_url = 7;
-
-    // Matches the Instrumentation Scope Schema URL.
-    bool scope_schema_url = 8;
+    // Standard first-class OpenTelemetry Resource fields.
+    ResourceField resource_field = 6;
   }
 }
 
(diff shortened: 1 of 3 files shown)


proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -18,11 +18,108 @@ package google.telemetry.policy.v1alpha1;
...
+
+    // Exact string equality match.
+    string exact = 11;
+
+    // Regular expression match using RE2 syntax.
+    string regex = 12;
+  }
+
+  // If true, inverts the result of the predicate (logical NOT).
+  bool negate = 20;
```
dashpole: Tags 1-15 encode in 1 byte on the wire while tag 20 takes 2 bytes. Since tags 2-9 are unused in `LogMatcher`, consider using a tag under 16 (like 2) to save a byte per matcher.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -18,11 +18,108 @@ package google.telemetry.policy.v1alpha1;
...
+    // Resource attribute by key (e.g. "service.name", "k8s.pod.name").
+    string resource_attribute = 3;
+
+    // Instrumentation scope attribute by key (e.g. "library.name").
+    string scope_attribute = 4;
+
+    // Matches the Instrumentation Scope name.
+    // (-- api-linter: core::0122::name-suffix=disabled
+    //     aip.dev/not-precedent: Matches OpenTelemetry InstrumentationScope name. --)
+    bool scope_name = 5;
```
dashpole: Using `bool` for `scope_name` (and the other scope/schema fields) inside `oneof target` allows an ambiguous `false` state and triggered the `core::0122` linter disable. Consider grouping scope and resource fields into dedicated enums (similar to `LogRecordField`) or using `google.protobuf.Empty`.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -18,11 +18,108 @@ package google.telemetry.policy.v1alpha1;
...
+  LOG_RECORD_FIELD_UNSPECIFIED = 0;
+
+  // The primary message body of the log record.
+  LOG_RECORD_FIELD_BODY = 1;
+
+  // The string severity of the log (e.g. "INFO", "ERROR", "DEBUG").
+  LOG_RECORD_FIELD_SEVERITY_TEXT = 2;
+
+  // The numerical severity of the log (1-24 per OpenTelemetry specification).
+  LOG_RECORD_FIELD_SEVERITY_NUMBER = 3;
```
dashpole: Since `SEVERITY_NUMBER` is an integer but `exact` and `regex` are string predicates, it would be helpful to document expected string coercion (e.g. matching against decimal digits `17` vs enum names). Clarifying whether `TRACE_ID` and `SPAN_ID` match lowercase hex strings versus raw bytes would also prevent evaluator discrepancies.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,136 @@ syntax = "proto3";
...
-// package hierarchy, repository organization, and xDS transport plumbing.
-message LogFilterPolicy {}
+// Policies are declarative, atomic, and transport-agnostic.
+message LogFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. "drop-noisy-healthchecks", "retain-error-logs").
+  string id = 1;
+
+  // The filtering action to take when all `matches` conditions evaluate to true.
+  Action action = 2;
```
dashpole: To follow Protobuf best practices (AIP-203) and allow validators and control planes to distinguish between an omitted action and `ACTION_UNSPECIFIED`, consider making `action` explicitly `optional`:
```suggestion
  optional Action action = 2;
```

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,136 @@ syntax = "proto3";
...
+// log records based on structured field and attribute matches.
 //
-// Note: Detailed filter conditions, matchers, and action semantics are deferred
-// to a subsequent design. This message is currently stubbed to establish the
-// package hierarchy, repository organization, and xDS transport plumbing.
-message LogFilterPolicy {}
+// Policies are declarative, atomic, and transport-agnostic.
+message LogFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. "drop-noisy-healthchecks", "retain-error-logs").
```
jsuereth: Nit: We should clarify if this is going to be a UUID.  If so (and I think the answer is LIKELY yes) we may want this to have a namespace, e.g. "io.opentelemetry.operator/{namespace}/policies/retain-error-logs" or some such.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,136 @@ syntax = "proto3";
...
+  }
+
+  // If true, inverts the result of the predicate (logical NOT).
+  bool negate = 2;
+}
+
+// LogFieldSelector designates a first-class OTel log field, attribute, or metadata.
+message LogFieldSelector {
+  oneof target {
+    // Standard first-class OpenTelemetry LogRecord fields.
```
jsuereth: Nit: In these descriptons, can you use the OTLP path?

I think aligning to the OTLP data model should make it easier for compliance tests later.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,136 @@ syntax = "proto3";
...
+    string resource_attribute = 3;
+
+    // Instrumentation scope attribute by key (e.g. "library.name").
+    string scope_attribute = 4;
+
+    // Standard first-class OpenTelemetry InstrumentationScope fields.
+    ScopeField scope_field = 5;
+
+    // Standard first-class OpenTelemetry Resource fields.
+    ResourceField resource_field = 6;
```
jsuereth: Nit: We are planning to deprecate resource.schema_url - maybe we don't add this at all?

jsuereth (approved):
The deviations from the OTEP make sense to me, and I think prove out with some of our prototyping as good changes. 

Let's talk with Jacob and others interested policy project and make sure these are tracked upstream.

Excited to see this experimental support start to land here!

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,136 @@ syntax = "proto3";
...
-// Note: Detailed filter conditions, matchers, and action semantics are deferred
-// to a subsequent design. This message is currently stubbed to establish the
-// package hierarchy, repository organization, and xDS transport plumbing.
-message LogFilterPolicy {}
+// Policies are declarative, atomic, and transport-agnostic.
+message LogFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. "drop-noisy-healthchecks", "retain-error-logs").
+  string id = 1;
+
```
rafaelwestphal, line 31: One thing we have been discussing is having a revision/version field in the policy itself. I believe it is better to make the policy agnostic of how they are ingested, so we can' t rely on the any transport protocol to keep track of the revision. We want to be able to track it in file ingested policy as well. I' m not sure how much schema evolution the protobufs can shield us from.

dashpole: I think that makes sense, but I need to think a bit more about alternatives.  Would you mind if we do that as a follow-up?

proto/policy/v1alpha1/log_filter_policy.proto
dashpole: addressed in bfd7f6b

proto/policy/v1alpha1/log_filter_policy.proto
dashpole: addressed in e29b1d6

proto/policy/v1alpha1/log_filter_policy.proto
dashpole: addressed in ab22413

proto/policy/v1alpha1/log_filter_policy.proto
dashpole: I'll tackle this after we have all 3 filter policies defined.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,124 @@ syntax = "proto3";
...
-// package hierarchy, repository organization, and xDS transport plumbing.
-message LogFilterPolicy {}
+// Policies are declarative, atomic, and transport-agnostic.
+message LogFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. a UUID or namespaced URI like "io.opentelemetry.operator/{namespace}/policies/retain-error-logs").
+  string id = 1;
+
+  // The filtering action to take when all `matches` conditions evaluate to true.
+  optional Action action = 2;
```
dashpole, line 33: PR #633 defined default allow and keep-overrides-drop precedence in `MetricFilterPolicy`. Should we document this in `common_policy_types.proto` or `log_filter_policy.proto` so log and trace policies share the same defined behavior?

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,124 @@ syntax = "proto3";
...
+// or attribute.
+message LogMatcher {
+  // The field, attribute, or metadata to inspect. Exactly one must be set.
+  LogFieldSelector target = 1;
+
+  // The comparison predicate to evaluate against the target value.
+  // Exactly one predicate must be set.
+  oneof predicate {
+    // Matches if the selected field or attribute exists.
+    google.protobuf.Empty exists = 10;
```
dashpole, line 52: For first-class fields like `record_field` and `scope_field` that lack explicit presence tracking in OTLP, it might be worth mentioning in docs that `exists` evaluates to true when the target is set to a non-default (non-zero/non-empty) value.

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,124 @@ syntax = "proto3";
...
+
+    // Exact string equality match.
+    string exact = 11;
+
+    // Regular expression match using RE2 syntax.
+    string regex = 12;
```
dashpole, line 58: PR #633 clarified for metrics that `regex` evaluates as an unanchored partial match. Should we backport that same clarification here and in trace filter policy?

proto/policy/v1alpha1/log_filter_policy.proto
```diff
@@ -16,13 +16,124 @@ syntax = "proto3";
...
+  bool negate = 2;
+}
+
+// LogFieldSelector designates a first-class OTel log field, attribute, or metadata.
+message LogFieldSelector {
+  oneof target {
+    // Standard first-class OpenTelemetry LogRecord fields (OTLP path: log_record.*).
+    LogRecordField record_field = 1;
+
+    // Log record attribute by key (OTLP path: log_record.attributes[*], e.g. "http.route", "user_id").
```
dashpole: `SEVERITY_NUMBER`, `TRACE_ID`, and `SPAN_ID` have explicit string coercion rules, but `BODY` and all attribute selectors resolve to OTLP `AnyValue` (int, bool, double, map). Consider specifying how `exact` and `regex` handle non-string values (e.g. whether primitive ints/bools are stringified or non-string values never match) so evaluators do not diverge on attributes like `http.response.status_code`.

dashpole (commented):
Follow-up TODOs.

PR #633, opened 2026-09-02
Add MetricFilterPolicy definition and generated Go code

dashpole:
## Summary

Following #632 (`LogFilterPolicy`) and #635 (`TraceFilterPolicy`), this PR adds **`MetricFilterPolicy`** under `google.telemetry.policy.v1alpha1`, along with generated Go bindings and unit tests.

It provides declarative filtering for OpenTelemetry metrics and individual metric data points, adhering to the common schema patterns established across the policy suite.

---

## Metric-Specific Schema Design

- **`MetricFieldSelector`**: Strongly typed selector designating:
  - **Metric Descriptor Fields**: `name`, `description`, `unit`, `type`, `aggregation_temporality`, and `is_monotonic` via `MetricDescriptorField` enum (OTLP path: `metric.*`).
  - **Data Point Attributes**: `datapoint_attribute` targeting attributes on individual metric data points (OTLP path: `metric.*.data_points[*].attributes[*]`, e.g. `http.status_code`, `env`).
  - **Resource & Scope Metadata**: `resource_attribute` (`resource.attributes[*]`), `scope_attribute` (`scope.attributes[*]`), and `ScopeField` (`scope.*`: `name`, `version`, `schema_url`). Note: `resource.schema_url` is omitted due to specification deprecation.
- **Universal Predicates**:
  - `exists` (`google.protobuf.Empty`), `exact` (string), and `regex` (RE2 syntax), with `negate` support.
  - Metric type is matched using `exact` (e.g. `"HISTOGRAM"`) or `regex` (e.g. `"HISTOGRAM|EXPONENTIAL_HISTOGRAM"`).
  - Aggregation temporality is matched using `exact` or `regex` against uppercase values (`"DELTA"`, `"CUMULATIVE"`).
  - Monotonicity is matched using `exact` against `"true"` or `"false"`.
- **Two-Level Evaluation Granularity**:
  - **Metric-Level Filtering**: If all matchers target metric-level fields (`descriptor_field`, `resource_attribute`, `scope_attribute`, `scope_field`), the action applies to the entire metric stream.
  - **DataPoint-Level Pruning**: If any matcher targets a `datapoint_attribute`, the action applies at the individual data point level—allowing users to drop noisy data points (e.g. `http.status_code == 200`) without discarding the entire metric stream.

---

## Differences with Upstream OTEP 4738 (with Justifications)

1. **DataPoint Pruning vs. Instrument Dropping**:
   - *OTEP 4738*: Evaluates policies at the metric instrument level (`bool keep`).
   - *This PR*: Supports two-level granularity based on whether `datapoint_attribute` is targeted.
   - *Justification*: Real-world cardinality control frequently requires filtering specific attribute combinations (e.g. dropping healthy status codes or internal endpoints) while retaining error signals on the same metric instrument. Dropping the entire metric instrument on a single data point match causes unacceptable telemetry loss.

2. **Metric Type, Temporality & Monotonicity via Descriptor Fields vs. Target-as-Value**:
   - *OTEP 4738*: Places `MetricType metric_type = 5;` inside `oneof field`, conflating field selection with target value and requiring dummy predicates. It omits temporality and monotonicity selectors entirely.
   - *This PR*: Promotes `TYPE`, `AGGREGATION_TEMPORALITY`, and `IS_MONOTONIC` into `MetricDescriptorField` evaluated via standard string predicates (`exact`, `regex`).
   - *Justification*: Eliminates the "target-as-value" anti-pattern, keeps `oneof predicate` 100% universal across all signal policies (`exists`, `exact`, `regex`), and allows expressive multi-type or multi-temporality matching in a single rule (e.g. `regex: "HISTOGRAM|GAUGE"`).

3. **`Action` Enum with Explicit Presence vs. `bool keep`**:
   - *OTEP 4738*: Uses `bool keep = 2;` (where default `false` acts as fail-closed drop).
   - *This PR*: Uses `optional Action action = 2;` (shared with `LogFilterPolicy` and `TraceFilterPolicy`) with `ACTION_UNSPECIFIED = 0` treated as invalid.
   - *Justification*: Enforces mandatory fail-open semantics (unconfigured or malformed policies cannot silently drop metrics). Marking it `optional` (AIP-203) provides explicit field presence tracking (`hasAction()`) so declarative validators and control planes can detect missing actions before runtime.

4. **`google.protobuf.Empty exists` vs. `bool exists`**:
   - *OTEP 4738*: Uses `bool exists = 12;` alongside `bool negate = 20;`.
   - *This PR*: Uses `google.protobuf.Empty exists = 10;` paired with `bool negate = 2;`.
   - *Justification*: Avoids the ambiguous `exists: false` state inside a `oneof` and eliminates contradictory double-negatives (`exists: false` + `negate: true`). Selecting `exists: {}` tests presence; `negate: true` tests absence.

5. **Structured `ScopeField` Enum vs. Loose Fields**:
   - *OTEP 4738*: Uses generic `AttributePath` for attributes and lacks explicit selectors for scope metadata.
   - *This PR*: Explicitly separates attributes from specification metadata (`ScopeField` for name, version, schema URL). `resource.schema_url` is omitted per OTel specification deprecation.
   - *Justification*: Strong typing prevents boolean flag clutter and clarifies that metadata fields are distinct from user-defined attributes.

---

## Alternatives Considered

1. **Split `MetricFilterPolicy` vs. `DataPointFilterPolicy` (Rejected)**
   - *Considered*: Defining separate protos for metric-level vs. data point-level filtering.
   - *Why Rejected*: Unnecessarily doubles the API surface. The presence of `datapoint_attribute` in the matchers unambiguously specifies data point evaluation granularity.
2. **Dedicated Enum Variant in `predicate` for `MetricType` (Rejected)**
   - *Considered*: Adding `MetricType metric_type` to `oneof predicate`.
   - *Why Rejected*: Leaks signal-specific enum types into the universal predicate oneof. Using `METRIC_DESCRIPTOR_FIELD_TYPE` with string predicates allows `MetricMatcher.predicate` to remain identical across all signals and supports regex matching.
3. **Boolean Presence Flags (`bool exists`, `bool scope_name`) (Rejected)**
   - *Considered*: Using `bool` fields inside `oneof target` and `oneof predicate`.
   - *Why Rejected*: In proto3, boolean `false` within a `oneof` is ambiguous and error-prone. Replaced with `ScopeField` enum and `google.protobuf.Empty`.

---

## Deferred Features

1. **Measurement / Value Thresholds** (e.g. `value > 100`): Requires numeric range operators (`gt`, `lt`) across int, double, and histogram distributions; deferred across all signal policies.
2. **Metric Views & Dimension Rollups**: Removing or aggregating away high-cardinality attributes is a transformation concern rather than a filter policy.


proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,145 @@
...
+    // Matches the Resource Schema URL.
+    bool resource_schema_url = 8;
+
+    // Matches the Instrumentation Scope Schema URL.
+    bool scope_schema_url = 9;
+  }
+}
+
+// Action specifies what to do with a metric or data point that matches all conditions.
+enum Action {
```
dashpole: `enum Action` is also defined at package level in #632 under `google.telemetry.policy.v1alpha1`, which causes `buf generate` to fail with duplicate symbol errors when both files are present. Maybe extract `Action` to a shared `action.proto` in this package, or nest it inside `MetricFilterPolicy`?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,145 @@
...
+  // Default unspecified action. Treated as ACTION_KEEP (fail-open).
```
dashpole: Notice `ACTION_UNSPECIFIED` is documented as fail-open here, whereas in #632 it is documented as invalid. We should probably align default action semantics across both policies.

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,155 @@
...
+    // Resource attribute by key (e.g. "service.name", "k8s.pod.name").
+    string resource_attribute = 4;
+
+    // Instrumentation scope attribute by key (e.g. "library.name").
+    string scope_attribute = 5;
+
+    // Matches the Instrumentation Scope name.
+    // (-- api-linter: core::0122::name-suffix=disabled
+    //     aip.dev/not-precedent: Matches OpenTelemetry InstrumentationScope name. --)
+    bool scope_name = 6;
```
dashpole: Using bool for scope_name, scope_version, and schema URLs in oneof target makes false an invalid or ambiguous state. Would it be cleaner to use google.protobuf.Empty or group scope fields into an enum (similar to MetricDescriptorField)?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,155 @@
...
+// MetricMatcher defines a single predicate evaluated against a specific metric
+// descriptor field, metric type, or data point attribute.
+message MetricMatcher {
+  // The metric field, type, or attribute to inspect. Exactly one must be set.
+  MetricFieldSelector target = 1;
+
+  // The comparison predicate to evaluate against the target value.
+  // Exactly one predicate must be set.
+  oneof predicate {
+    // Matches if the selected field or attribute exists.
```
dashpole: Having bool exists alongside bool negate allows contradictory combinations like exists: false with negate: true. Could exists use google.protobuf.Empty so presence is indicated by selecting the variant and absence by negate: true?

gen/go/policy/v1alpha1/metric_filter_policy_test.go
```diff
@@ -0,0 +1,192 @@
...
+			{
+				Target: &policyv1alpha1.MetricFieldSelector{
+					Target: &policyv1alpha1.MetricFieldSelector_DescriptorField{
+						DescriptorField: policyv1alpha1.MetricDescriptorField_METRIC_DESCRIPTOR_FIELD_NAME,
+					},
+				},
+				Predicate: &policyv1alpha1.MetricMatcher_Regex{
+					Regex: "^http\\.server\\..*",
+				},
+				Negate: false,
```
dashpole: None of the matchers currently exercise Negate: true. Could we add a test case with Negate: true to verify negative predicate matching and round-trip serialization?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,153 @@
...
+//
+// Evaluation Granularity:
+// - Instrument-level: If all matchers in `matches` target metric-level fields
+//   (descriptor_field, resource_attribute, scope_attribute,
+//   scope_field, or resource_field), the action applies to the entire metric
+//   stream (e.g. ACTION_DROP drops the entire metric).
+// - Data point-level: If any matcher in `matches` targets a
+//   `datapoint_attribute`, the action applies at the individual data point level
+//   (e.g. ACTION_DROP prunes matching data points while preserving non-matching
+//   data points in the metric stream).
```
dashpole: Per feedback on #632, we should clarify that `id` is typically a UUID or namespaced URI:
```suggestion
  // (e.g. a UUID or namespaced URI like "io.opentelemetry.operator/{namespace}/policies/drop-jvm-metrics").
```

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,153 @@
...
+
+    // Regular expression match using RE2 syntax.
+    string regex = 12;
+  }
+
+  // If true, inverts the result of the predicate (logical NOT).
+  bool negate = 2;
+}
+
+// MetricFieldSelector designates a first-class OTel metric field, attribute, or metadata.
```
dashpole: Per feedback on #632, can we include the OTLP data model paths in these field descriptions (e.g. `metric.*`, `metric.*.data_points[*].attributes[*]`, `resource.attributes[*]`, `scope.attributes[*]`, `scope.*`) to make compliance tests easier?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,153 @@
...
+    string datapoint_attribute = 2;
+
+    // Resource attribute by key (e.g. "service.name", "k8s.pod.name").
+    string resource_attribute = 3;
+
+    // Instrumentation scope attribute by key (e.g. "library.name").
+    string scope_attribute = 4;
+
+    // Standard first-class OpenTelemetry InstrumentationScope fields.
+    ScopeField scope_field = 5;
```
dashpole: Since `resource.schema_url` is being deprecated (and per feedback on #632), let's remove `resource_field` and the `ResourceField` enum.

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,115 @@
...
+    // Instrumentation scope attribute by key (OTLP path: scope.attributes[*], e.g. "library.name").
+    string scope_attribute = 4;
+
+    // Standard first-class OpenTelemetry InstrumentationScope fields (OTLP path: scope.*).
+    ScopeField scope_field = 5;
+  }
+}
+
+// MetricDescriptorField identifies top-level descriptor fields of an OpenTelemetry Metric.
+enum MetricDescriptorField {
```
jsuereth, line 111: Do we need something like this for fields *on* the metric types, e.g. DELTA vs. CUMULATIVE

jsuereth (approved):
Approving for similar reasons as the LogFilterPolicy - same changes, same rationale.

proto/policy/v1alpha1/metric_filter_policy.proto
dashpole: I have that listed as a deferred feature.  Would you rather it was included from the start?

dashpole: Added.  No point in waiting since we will need these eventually

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,124 @@
...
+// and attribute matches.
+//
+// Policies are declarative, atomic, and transport-agnostic.
+//
+// Evaluation Granularity:
+// - Instrument-level: If all matchers in `matches` target metric-level fields
+//   (descriptor_field, resource_attribute, scope_attribute, or scope_field),
+//   the action applies to the entire metric stream (e.g. ACTION_DROP drops
+//   the entire metric).
+// - Data point-level: If any matcher in `matches` targets a
```
dashpole: Two edge cases in the data-point evaluation docs: (a) if an ACTION_DROP policy prunes all data points of a metric, is the empty metric emitted or dropped completely? (b) What does ACTION_KEEP mean at data-point level—is it an allowlist that prunes non-matching data points, or an exemption?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,124 @@
...
+  // The comparison predicate to evaluate against the target value.
+  // Exactly one predicate must be set.
+  oneof predicate {
+    // Matches if the selected field or attribute exists.
+    google.protobuf.Empty exists = 10;
+
+    // Exact string equality match.
+    string exact = 11;
+
+    // Regular expression match using RE2 syntax.
```
dashpole: Does regex perform a full match (like Envoy RegexMatcher) or an unanchored partial match (like Go regexp.MatchString)? Worth clarifying whether callers need explicit anchors.

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,124 @@
...
+  bool negate = 2;
+}
+
+// MetricFieldSelector designates a first-class OTel metric field, attribute, or metadata.
+message MetricFieldSelector {
+  oneof target {
+    // Top-level metric descriptor fields (OTLP path: metric.*).
+    MetricDescriptorField descriptor_field = 1;
+
+    // Attribute on an individual metric data point (OTLP path: metric.*.data_points[*].attributes[*], e.g. "http.status_code", "env").
```
dashpole: For non-string AnyValue attributes like http.status_code (int) or flags (bool), do exact and regex evaluate against canonical string representations (e.g. "200", "true"), or are non-string attributes unmatchable?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,137 @@
...
+    string datapoint_attribute = 2;
+
+    // Resource attribute by key (OTLP path: resource.attributes[*], e.g. "service.name", "k8s.pod.name").
+    string resource_attribute = 3;
```
dashpole: `Action`, `ScopeField` and `ResourceField` are package-scoped, so these collide with the identical declarations in #632 and #635 — twelve duplicated type and value names in `google.telemetry.policy.v1alpha1`, and all three generated `.pb.go` files declare `type Action int32` in package `policyv1alpha1`. This PR is green against master today, but the second of the three to merge is a hard proto and Go compile failure, so `make lint-protos` / `make compare-protos` break. Since none of these three enums is metric-specific, and the `exists`/`exact`/`regex` + `negate` block at `:61–73` is byte-identical across all three files, could they move into a shared `policy/v1alpha1/common.proto`?

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,137 @@
...
+import "policy/v1alpha1/common_policy_types.proto";
+
+option go_package = "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/gen/go/policy/v1alpha1;policyv1alpha1";
+
+// MetricFilterPolicy defines an intent-based rule for retaining or dropping
+// metrics or individual metric data points based on structured field, type,
+// and attribute matches.
+//
+// Policies are declarative, atomic, and transport-agnostic.
+//
```
dashpole: The granularity *selection* rule at `:30`/`:34` is unambiguous, including the mixed case — but the outcome isn't. Two gaps: (a) if every data point of a metric matches an `ACTION_DROP` policy, is an empty metric emitted or is the metric dropped? `:36–37`'s "preserving non-matching data points in the metric stream" reads either way with zero survivors, and the two are observably different downstream. (b) Both bullets only exemplify `ACTION_DROP`; at data-point granularity `ACTION_KEEP` is either a no-op or an allow-list that prunes every non-matching data point, and nothing here picks one.

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,137 @@
...
+  // to true for the policy action to take effect.
+  // At least one matcher is required; a policy with no matchers is invalid.
+  repeated MetricMatcher matches = 3;
+}
+
+// MetricMatcher defines a single predicate evaluated against a specific metric
+// descriptor field or data point attribute.
+message MetricMatcher {
+  // The metric field, attribute, or metadata to inspect. Exactly one must be set.
+  MetricFieldSelector target = 1;
```
dashpole: Same two gaps as #632 (F3/F4), so one answer covers both — "RE2 syntax" doesn't pick anchored vs unanchored (Envoy's `RegexMatcher` is full-match, OTTL's `IsMatch` is not, and both call themselves RE2), and `datapoint_attribute` / `resource_attribute` / `scope_attribute` all resolve to `AnyValue` with no stated coercion, so `{datapoint_attribute: "http.status_code", exact: "404"}` — this file's own example attribute, an int under semconv — matches or silently never matches depending on the implementation. Worth noting that #632 at least pinned coercion for `SEVERITY_NUMBER`/`TRACE_ID`/`SPAN_ID`; this file has no coercion rule at all.

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,137 @@
...
+  oneof target {
+    // Top-level metric descriptor fields (OTLP path: metric.*).
+    MetricDescriptorField descriptor_field = 1;
+
+    // Attribute on an individual metric data point (OTLP path: metric.*.data_points[*].attributes[*], e.g. "http.status_code", "env").
+    string datapoint_attribute = 2;
+
+    // Resource attribute by key (OTLP path: resource.attributes[*], e.g. "service.name", "k8s.pod.name").
+    string resource_attribute = 3;
+
```
dashpole: `:101` says an unspecified action makes the policy invalid, but #635 documents the *same package-scoped* `ACTION_UNSPECIFIED` as "Treated as ACTION_KEEP (fail-open)". Only one of these survives once the duplicate enums are reconciled (F1), and picking by accident silently changes one signal's documented behavior — worth choosing one rule for all three signals now, and saying what an implementation actually does with an invalid policy.

proto/policy/v1alpha1/metric_filter_policy.proto
```diff
@@ -0,0 +1,137 @@
...
+  METRIC_DESCRIPTOR_FIELD_NAME = 1;
+
+  // The descriptive text explaining the metric (OTLP path: metric.description).
+  METRIC_DESCRIPTOR_FIELD_DESCRIPTION = 2;
+
+  // The unit of measure (OTLP path: metric.unit, e.g. "ms", "By", "1").
+  METRIC_DESCRIPTOR_FIELD_UNIT = 3;
+
+  // The primary data type of the metric (e.g. "GAUGE", "SUM", "HISTOGRAM",
+  // "EXPONENTIAL_HISTOGRAM", "SUMMARY"). When selected, string predicates
```
dashpole: `:125–128` is close to a complete mapping but is introduced with `e.g.`, and "uppercase type names" isn't a mechanical transform of anything — `strings.ToUpper("ExponentialHistogram")` is `EXPONENTIALHISTOGRAM`, not the `EXPONENTIAL_HISTOGRAM` in the list, so the word-boundary rule lives only in the examples. The `Metric.data`-unset case (pdata's `MetricTypeEmpty`) isn't listed at all. Since this is the one place a typed value becomes a wire-visible string, worth enumerating it normatively per type and saying the SCREAMING_SNAKE form is deliberate (it matches neither `pmetric.MetricType.String()` nor the OTLP JSON names).

menderico:
Late here, but LGTM.

PR #635, opened 2026-09-02
Add TraceFilterPolicy protobuf definition and Go bindings

dashpole:
Following the Log filter policy definition established in #632, this PR introduces the canonical **`TraceFilterPolicy`** schema and generated Go bindings.

## Trace-Specific Schema Summary

- **`SpanRecordField`**: Selects first-class span fields (`OTLP path: span.*`):
  - `NAME`: Span operation name (e.g. `HTTP GET /users`).
  - `TRACE_ID`, `SPAN_ID`, `PARENT_SPAN_ID`: 32-character (trace) and 16-character (span) lowercase hex strings matching W3C TraceContext format.
  - `STATUS_MESSAGE`: Error status description.
  - `KIND`: Matches uppercase span kind names (`"INTERNAL"`, `"SERVER"`, `"CLIENT"`, `"PRODUCER"`, `"CONSUMER"`).
  - `STATUS_CODE`: Matches uppercase status code names (`"OK"`, `"ERROR"`, `"UNSET"`).
- **`span_attribute`**: Selects span attributes by string key (`OTLP path: span.attributes[*]`, e.g. `http.route`, `db.system`, `rpc.method`).
- **Universal Predicates**: Evaluated using standard `exists` (`google.protobuf.Empty`), `exact`, and `regex` predicates with `negate` support, maintaining schema consistency with `LogMatcher` and `MetricMatcher`.
- **Shared Types**: `Action` (`optional Action action = 2`) and `ScopeField` are defined in `common_policy_types.proto`, shared across `LogFilterPolicy` and `TraceFilterPolicy`.

---

## Comparison with Upstream OTEP 4738

- **Stateless Span Filtering vs. Probabilistic Trace Sampling**:
  - *OTEP 4738*: `TraceTarget` only specifies probabilistic head sampling (`TraceSamplingConfig` with `percentage`, `mode`, `hash_seed`, `tracestate`). It does not define deterministic span keep/drop filtering.
  - *This PR*: Defines deterministic, stateless boolean span filtering (`KEEP` vs `DROP`). Probabilistic sampling is deferred to a dedicated `TraceSamplingPolicy`.
  - *Justification*: Telemetry filtering (e.g. dropping high-volume health checks or retaining errors) is an orthogonal concern from statistical sampling across distributed traces.
- **Span Kind & Status Code Matching**:
  - *OTEP 4738*: Placed `SpanKind span_kind = 5` and `SpanStatusCode span_status = 6` in `oneof field`, inverting the selector/predicate model (target held the expected value, requiring a dummy `exists: true` predicate).
  - *This PR*: `SPAN_RECORD_FIELD_KIND` and `SPAN_RECORD_FIELD_STATUS_CODE` are selected in `SpanRecordField`, and evaluated via universal string predicates (`exact`, `regex`).
  - *Justification*: Keeps `TraceMatcher.predicate` strictly identical across logs, metrics, and traces, and enables regex matching across multiple values (e.g. `regex: "SERVER|CONSUMER"` or `regex: "ERROR|UNSET"`).
- **Trace/Span ID Hex String Formatting & `PARENT_SPAN_ID`**:
  - *OTEP 4738*: Referenced raw trace fields without defining string representations or including `parent_span_id`.
  - *This PR*: Explicitly documents 32-char and 16-char lowercase hex string matching per W3C TraceContext, and includes `SPAN_RECORD_FIELD_PARENT_SPAN_ID` to allow targeting root spans or specific span subtrees.
  - *Justification*: Eliminates ambiguity for string matchers against byte fields and supports root-span identification.
- **Deferred Span Events & Links**:
  - *OTEP 4738*: Included `event_name`, `event_attribute`, and `link_trace_id` in `TraceMatcher`.
  - *This PR*: Deferred to follow-up iterations.
  - *Justification*: Span events and links are nested repeated collections requiring iteration and quantifier semantics ("any" vs "all"). Deferring them keeps the span evaluation path fast and zero-allocation; they can be added cleanly as target variants in the future.

---

## Alternatives Considered

1. **`SpanFilterPolicy` Naming (Rejected)**: Named `TraceFilterPolicy` to maintain signal-level consistency (`LogFilterPolicy`, `MetricFilterPolicy`) and align with Google Cloud Telemetry Management APIs (`collectors.proto`).
2. **Dedicated Enum Predicates in `TraceMatcher` (Rejected)**: Considered adding `SpanKind span_kind` and `SpanStatusCode span_status` to `oneof predicate`. Rejected to preserve a universal predicate schema across all signals and enable multi-value regex matches.
3. **Stateful Tail-Based Trace Assembly (Rejected)**: Reconstructing complete distributed trace trees requires buffering and memory overhead. Tail sampling belongs in dedicated collector processors; filter policies remain fast, stateless, and evaluated per span.

---

## Deferred Features

1. **Span Events and Links**: Can be added as dedicated target variants in `TraceFieldSelector` without breaking wire compatibility.
2. **Probabilistic Sampling & `tracestate`**: Deferred to `TraceSamplingPolicy`.
3. **Duration / Latency Thresholds (e.g. `duration > 500ms`)**: Numeric comparisons are deferred across all filter policies.


proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,166 @@
...
+message TraceFieldSelector {
+  oneof target {
+    // Standard first-class OpenTelemetry Span fields (name, trace/span IDs, status message).
+    SpanRecordField record_field = 1;
+
+    // Span attribute by key (e.g. "http.route", "db.system", "rpc.method").
+    string span_attribute = 2;
+
+    // Span Kind. Setting this variant defines the target enum; use predicate
+    // exists=true (or negate=true) to match.
```
dashpole: Putting the expected enum in `target` while requiring a dummy `exists: true` predicate inverts the selector/predicate model. Could `span_kind` and `span_status` be matching variants in `predicate`, with `target` simply selecting the field?

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,166 @@
...
+// attribute, or status.
+message TraceMatcher {
+  // The span field, attribute, or metadata to inspect. Exactly one must be set.
+  TraceFieldSelector target = 1;
+
+  // The comparison predicate to evaluate against the target value.
+  // Exactly one predicate must be set.
+  oneof predicate {
+    // Matches if the selected field or attribute exists.
+    bool exists = 10;
```
dashpole: Having `bool exists` alongside `bool negate` creates ambiguous combinations like `exists: false` with `negate: true`. If `negate` handles inversion, maybe `exists` could be `google.protobuf.Empty` or documented to only accept `true`?

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,166 @@
...
+// SpanRecordField identifies standard first-class fields of an OpenTelemetry Span.
+enum SpanRecordField {
+  // Default unspecified span record field.
+  SPAN_RECORD_FIELD_UNSPECIFIED = 0;
+
+  // The operation name of the span (e.g. "HTTP GET /users", "SELECT * FROM users").
+  SPAN_RECORD_FIELD_NAME = 1;
+
+  // The 16-byte trace identifier associated with the span.
+  SPAN_RECORD_FIELD_TRACE_ID = 2;
```
dashpole: OTel stores trace and span IDs as raw bytes, while `exact` and `regex` take strings. Worth clarifying in the doc comment whether these match against lowercase hex strings (32-char trace ID, 16-char span ID)?

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,202 @@
...
+// spans based on structured field, attribute, and status matches.
+//
+// Policies are declarative, atomic, and transport-agnostic.
+message TraceFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. "drop-healthcheck-spans", "retain-error-spans").
+  string id = 1;
+
+  // The filtering action to take when all `matches` conditions evaluate to true.
+  Action action = 2;
```
dashpole: To follow Protobuf best practices (AIP-203) and allow declarative validators to distinguish an omitted action from `ACTION_UNSPECIFIED`, consider making `action` explicitly `optional`:
```suggestion
  optional Action action = 2;
```

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,202 @@
...
+    string exact = 11;
+
+    // Regular expression match using RE2 syntax.
+    string regex = 12;
+
+    // Matches the span kind.
+    SpanKind span_kind = 13;
+
+    // Matches the span status code.
+    SpanStatusCode span_status = 14;
```
dashpole: To keep `TraceMatcher.predicate` universal across all three signal policies (matching `LogMatcher` and `MetricMatcher` with only `exists`, `exact`, and `regex`), consider removing `span_kind` and `span_status` from `oneof predicate` and matching `SPAN_RECORD_FIELD_KIND` and `SPAN_RECORD_FIELD_STATUS_CODE` using standard string predicates (`exact: "SERVER"`, `regex: "SERVER|CLIENT"`, `exact: "ERROR"`).

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,163 @@
...
+
+option go_package = "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/gen/go/policy/v1alpha1;policyv1alpha1";
+
+// TraceFilterPolicy defines an intent-based rule for retaining or dropping
+// spans based on structured field, attribute, and status matches.
+//
+// Policies are declarative, atomic, and transport-agnostic.
+message TraceFilterPolicy {
+  // Unique identifier for the policy within its configuration scope
+  // (e.g. "drop-healthcheck-spans", "retain-error-spans").
```
dashpole: Per feedback on #632, we should clarify that `id` is typically a UUID or namespaced URI:
```suggestion
  // (e.g. a UUID or namespaced URI like "io.opentelemetry.operator/{namespace}/policies/retain-error-spans").
```

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  // If true, inverts the result of the predicate (logical NOT).
+  bool negate = 2;
+}
+
+// TraceFieldSelector designates a first-class OTel span field, attribute, or metadata.
+message TraceFieldSelector {
+  oneof target {
+    // Standard first-class OpenTelemetry Span fields (name, trace/span IDs, status message).
```
dashpole: Per feedback on #632, can we include the OTLP data model paths in these field descriptions (e.g. `span.*`, `span.attributes[*]`, `resource.attributes[*]`, `scope.attributes[*]`, `scope.*`) to make compliance tests easier?

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,163 @@
...
+    string resource_attribute = 3;
+
+    // Instrumentation scope attribute by key (e.g. "library.name").
+    string scope_attribute = 4;
+
+    // Standard first-class OpenTelemetry InstrumentationScope fields.
+    ScopeField scope_field = 5;
+
+    // Standard first-class OpenTelemetry Resource fields.
+    ResourceField resource_field = 6;
```
dashpole: Since `resource.schema_url` is being deprecated (and per feedback on #632), let's remove `resource_field` and the `ResourceField` enum.

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+// Action specifies what to do with a span that matches all conditions.
+enum Action {
+  // Default unspecified action. Treated as ACTION_KEEP (fail-open).
```
dashpole: To match #632 and #633, let's clarify that an unspecified action is invalid rather than treated as `ACTION_KEEP`:
```suggestion
  // Default unspecified action. A policy with an unspecified action is invalid.
```

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package google.telemetry.policy.v1alpha1;
+
+import "google/protobuf/empty.proto";
+import "policy/v1alpha1/log_filter_policy.proto";
```
jsuereth: Nit - if you're sharing types, move them into `common_policy_types.proto`

jsuereth (approved):
I think this makes sense, but my overall concern here is the interaction between this an tail-sampling.

I.e. Trace filtering is more of a "sampling" decision sometimes, but there *are* use cases to remove "useless" spans, e.g. if instrumentation all HTTP traffic, calls to `/metrics` endpoint may be something I drop.

That's a much larger discussion - let's keep this proposal aligned to the changes we want to propose from the OTEP.

proto/policy/v1alpha1/trace_filter_policy.proto
dashpole: addressed in cc1c151

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+import "google/protobuf/empty.proto";
+import "policy/v1alpha1/common_policy_types.proto";
+
+option go_package = "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/gen/go/policy/v1alpha1;policyv1alpha1";
+
+// TraceFilterPolicy defines an intent-based rule for retaining or dropping
+// spans based on structured field, attribute, and status matches.
+//
+// Policies are declarative, atomic, and transport-agnostic.
+message TraceFilterPolicy {
```
dashpole, line 28: The PR description settles this — "filter policies remain fast, stateless, and evaluated per span", with trace assembly explicitly rejected — but none of it is in the `.proto`, where line 26's "declarative, atomic, and transport-agnostic" leaves "atomic" undefined.

PR #633 wrote an explicit `Evaluation Granularity:` block into `metric_filter_policy.proto`. Two sentences here — that spans are evaluated independently without reassembly, and that dropping a span does not drop its descendants — would match the sibling and warn policy authors that dropping a span in the middle of a trace leaves its child spans behind as orphans with a dangling `parent_span_id`.

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+
+  // The 8-byte span identifier.
+  // When evaluated against string predicates (`exact`, `regex`), this matches
+  // against the 16-character lowercase hexadecimal string representation
+  // per the OpenTelemetry / W3C Trace Context specification.
+  SPAN_RECORD_FIELD_SPAN_ID = 3;
+
+  // The 8-byte parent span identifier.
+  // When evaluated against string predicates (`exact`, `regex`), this matches
+  // against the 16-character lowercase hexadecimal string representation
```
dashpole, line 108: `TRACE_ID`, `SPAN_ID`, and `STATUS_CODE` are pinned precisely, but default values for `PARENT_SPAN_ID` and `KIND` are left open:

1. **Root spans / absent `parent_span_id`**: A root span's parent ID is empty. `pcommon.SpanID.String()` renders `""` (and pdata's JSON encoder omits the field entirely), while `hex.EncodeToString` renders `"0000000000000000"`. If treated as the latter, `{record_field: PARENT_SPAN_ID, exists: {}}` evaluates to true for *every* span and drops root spans along with children. Since root-span targeting is an advertised use case, it would be good to specify explicitly whether `exists` evaluates to false for a root span (or what string value it matches).
2. **`KIND`**: OTLP says implementations MAY treat `SPAN_KIND_UNSPECIFIED` as `INTERNAL`. It would be helpful to clarify whether an unset kind matches `exact: "INTERNAL"`, and document the prefix-stripping rule (like `STATUS_CODE` has for `STATUS_CODE_`).

dashpole (commented):
Some follow-up TODOs.

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+
+    // Standard first-class OpenTelemetry InstrumentationScope fields (OTLP path: scope.*).
+    ScopeField scope_field = 5;
+  }
+}
+
+// SpanRecordField identifies standard first-class fields of an OpenTelemetry Span.
+enum SpanRecordField {
+  // Default unspecified span record field.
+  SPAN_RECORD_FIELD_UNSPECIFIED = 0;
```
dashpole: `Action`, `ScopeField` and `ResourceField` are package-scoped, and #632 and #633 declare identical copies in `google.telemetry.policy.v1alpha1` — `buf build` over the three branches gives 12 `declared multiple times` errors, and the three `.pb.go` files each declare `type Action int32` in the same Go package, so `go build` gives `Action redeclared in this block`. Whichever of the three merges second fails `make lint-protos` and `make lint`. (This PR is clean against master on its own, since master has only the empty `LogFilterPolicy` stub.) The useful thing now is to name the destination while #632 is still open: a `policy/v1alpha1/common.proto` for these three enums plus the `exists`/`exact`/`regex` + `negate` block, which is byte-identical in all three files. Otherwise the default outcome is that the trace and metric policies import `log_filter_policy.proto` permanently.

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+
+  // The status message describing an error.
+  SPAN_RECORD_FIELD_STATUS_MESSAGE = 5;
+
+  // The span kind (e.g. "INTERNAL", "SERVER", "CLIENT", "PRODUCER", "CONSUMER").
+  // When evaluated against string predicates (`exact`, `regex`), this matches
+  // against uppercase span kind string names.
+  SPAN_RECORD_FIELD_KIND = 6;
+
+  // The span status code (e.g. "OK", "ERROR", "UNSET").
```
dashpole: `TRACE_ID`/`SPAN_ID`/`STATUS_CODE` are pinned precisely, but the two fields whose *default* value is normal rather than degenerate are not. A root span's `parent_span_id` is empty: `pcommon.SpanID.String()` renders that `""` (and pdata's JSON encoder omits the field entirely — `internal/generated_proto_span.go:194`) while `hex.EncodeToString` renders `"0000000000000000"`, which is what `:120–124`'s "16-character lowercase hexadecimal" wording endorses. So `{record_field: PARENT_SPAN_ID, exists: {}}` is either false for root spans or true for every span — the second reading drops the whole trace. Since the description gives root-span targeting as this field's purpose, worth saying explicitly whether `exists` is false for a root span. Same shape on `:129`: `KIND` lists five of six values with no prefix rule, and OTLP says implementations MAY treat `SPAN_KIND_UNSPECIFIED` as `INTERNAL`, so an unset kind matches `exact: "INTERNAL"` on some implementations and not others.

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+
+syntax = "proto3";
+
+package google.telemetry.policy.v1alpha1;
+
+import "google/protobuf/empty.proto";
+import "policy/v1alpha1/common_policy_types.proto";
+
+option go_package = "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/gen/go/policy/v1alpha1;policyv1alpha1";
+
```
dashpole: The PR description settles this — "filter policies remain fast, stateless, and evaluated per span", with trace assembly explicitly rejected — but none of it is in the `.proto`, where `:26`'s "declarative, atomic, and transport-agnostic" leaves "atomic" undefined. #633 wrote exactly this kind of rule into `metric_filter_policy.proto:29–37` as an `Evaluation Granularity:` block; the trace file, where the unit matters most, has no equivalent. Two sentences here — spans evaluated independently, no reassembly, and dropping a span does not drop its descendants — would match the sibling and would also warn policy authors that `{action: ACTION_DROP, matches: [{span_attribute: "http.route", exact: "/healthz"}]}` leaves the healthcheck's child spans behind as orphans with a dangling `parent_span_id`.

proto/policy/v1alpha1/trace_filter_policy.proto
```diff
@@ -0,0 +1,125 @@
...
+    // Standard first-class OpenTelemetry InstrumentationScope fields (OTLP path: scope.*).
+    ScopeField scope_field = 5;
+  }
+}
+
+// SpanRecordField identifies standard first-class fields of an OpenTelemetry Span.
+enum SpanRecordField {
+  // Default unspecified span record field.
+  SPAN_RECORD_FIELD_UNSPECIFIED = 0;
+
```
dashpole: `:90` documents `ACTION_UNSPECIFIED` as "Treated as ACTION_KEEP (fail-open)", but this is a package-scoped enum value and both #632 (`:90`) and #633 (`:101`) document the same value as "A policy with an unspecified action is invalid" — the only substantive difference between the three `Action` enum bodies. Only one comment survives the de-dup in F1, and picking by accident silently changes two signals' documented behaviour. Worth choosing one rule for all three now; note that under the fail-open reading, `optional Action` at `:33` distinguishes nothing, since absent and explicit-`ACTION_UNSPECIFIED` both mean keep.

PR #643, opened 2026-09-04
pkg/googlepolicy and provider proof of concept

braydonk:
This PR contains the proof of concept for the googlecontrolplane provider design. Support for reading policy files from a directory is included, along with an example destination and source policy. Some other mechanisms like allowing processors to retrieve from a changed policy set, proper error handling and rollback, an extension to report metrics, event reporting, and more are still necessary to be demoable.


specs/gboc-policies.yaml
XuechunHou, line 1: Why is this file needed?

XuechunHou (approved):

pkg/googlepolicy/registry.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// ActivePolicySetRevisionID returns the revision ID of the active policy set.
+// Sometimes the only thing we need to know about the active policy set is the
+// revision ID, so rather than forcing callers to get a reference to the entire
+// set, they can just get a string.
+func ActivePolicySetRevisionID() string {
+	policySetMu.RLock()
+	defer policySetMu.RUnlock()
+	return activePolicySet.RevisionID
```
dashpole, line 90: If `activePolicySet` is nil (such as initially or after rolling back to empty history), this will panic with a nil pointer dereference. We should return an empty string if `activePolicySet == nil`.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/selfmetrics/selfmetrics_policy.go
```diff
@@ -0,0 +1,206 @@
...
+func (p *SelfMetricsPolicy) PolicyType() string {
+	return PolicyType
+}
+
+func (p *SelfMetricsPolicy) PolicyClass() googlepolicy.PolicyClass {
+	return googlepolicy.PolicyClassSource
+}
+
+func (p *SelfMetricsPolicy) Validate() error {
+	panic("unimplemented")
```
dashpole: Calling `Validate` here panics with unimplemented, which causes `googlepolicy.LoadPolicy` to panic whenever a `self_metrics` policy is loaded from config. Should this validate that `p.Name != ""` like `GCPDestinationPolicy` does?
```suggestion
	if p.Name == "" {
		return errors.New("policy must be named")
	}
	return nil
```

pkg/googlepolicy/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy
+
+go 1.26.6
+
+require (
+	github.com/stretchr/testify v1.12.1
+	go.opentelemetry.io/collector/confmap v1.66.0
```
dashpole: `policy.go` imports `go.opentelemetry.io/collector/component` for `component.ID`, but it is missing from `go.mod`, causing `go test` in this module to fail. Running `go mod tidy` in `pkg/googlepolicy` resolves this.

components/google-built-opentelemetry-collector/registry.yaml
```diff
@@ -17,10 +17,20 @@ exporters:
     gomod: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter v0.0.0
     path: "../components/google-built-opentelemetry-collector/exporter/googleservicecontrolexporter"
 
+processors:
+  googlepolicy:
+    gomod: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/components/google-built-opentelemetry-collector/processor/googlepolicy v0.0.0
+    path: "../components/google-built-opentelemetry-collector/provider/googlecontrolplane"
```
dashpole: The `gomod` and `path` here point to `.../processor/googlepolicy` and `../components/.../provider/googlecontrolplane`, but the processor module is named `googlepolicyprocessor` and lives in `../components/google-built-opentelemetry-collector/processor/googlepolicyprocessor`:
```suggestion
    path: "../components/google-built-opentelemetry-collector/processor/googlepolicyprocessor"
```
Also remember to update the `gomod` line to end with `googlepolicyprocessor`.

components/google-built-opentelemetry-collector/registry.yaml
```diff
@@ -17,10 +17,20 @@ exporters:
...
   oauth2clientauth:
     gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/oauth2clientauthextension v0.0.0
-    path: "../components/google-built-opentelemetry-collector/extension/oauth2clientauthextension"
\ No newline at end of file
+    path: "../components/google-built-opentelemetry-collector/extension/oauth2clientauthextension"
+
+providers:
+  googlecontrolplane:
+    gomod: github.com/GoogleCloudPlatform/opentelemetry-operations-collector/components/google-built-opentelemetry-collector/provider/googlecontrolplane v0.0.0
+    path: "../components/google-built-opentelemetry-collector/provider/googlecontrolplane"
```
dashpole: The provider directory is named `googlecontrolplaneprovider` on disk, so this path currently points to a missing directory:
```suggestion
    path: "../components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider"
```

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/gcpdestination/gcpdestination_policy.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+	}
+	otlpExporterType, _ := component.NewType("otlp_grpc")
+	otlpExporterID := component.NewIDWithName(otlpExporterType, p.Name)
+
+	conf.Exporters = map[component.ID]component.Config{
+		otlpExporterID: component.Config(otlpExporter),
+	}
+
+	queueBatchType, _ := component.NewType("queue_batch")
```
dashpole: The component type for the queuebatch processor is `queuebatch` without an underscore (defined as `queuebatchprocessor.NewFactory().Type()`). Hardcoding `queue_batch` here will prevent the collector from resolving the processor at startup:
```suggestion
	queueBatchType, _ := component.NewType("queuebatch")
```

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/file_policy_manager.go
```diff
@@ -0,0 +1,241 @@
...
+		defer fpm.wg.Done()
+		for {
+			select {
+			case <-fpm.done:
+				return
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				if strings.EqualFold(filepath.Ext(event.Name), ".json") && (event.Has(fsnotify.Write) || event.Has(fsnotify.Create)) {
```
dashpole, line 115: Should we check for `fsnotify.Rename` or `fsnotify.Remove` as well so removed or atomically replaced policy files trigger a reload?

dashpole:
any idea why presubmits didn't run?

braydonk:
I'm assuming the `checkout` action has some restriction if it finds there's conflicts on the branch? I'm gonna fix the conflicts and see if they actually run now.

specs/gboc-policies.yaml
braydonk: it's not, got left over in the branch and can be deleted

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+func newProvider(set confmap.ProviderSettings) confmap.Provider {
+	return &provider{
+		logger: set.Logger,
+	}
+}
+
+// Retrieve retrieves the configuration from the Google Control Plane provider for the given URI.
+func (p *provider) Retrieve(ctx context.Context, uri string, watcher confmap.WatcherFunc) (*confmap.Retrieved, error) {
```
dashpole, line 89: `watcher confmap.WatcherFunc` is accepted here by `Retrieve`, but it is currently discarded and never called when policy updates arrive. Consider invoking `watcher(&confmap.ChangeEvent{})` when destination or source policies change so the collector runtime triggers a configuration reload and reconstructs the pipelines. We should retain `watcher` on the provider instance (or pass it through to the policy manager) to drive that reload loop.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
```diff
@@ -0,0 +1,325 @@
...
+			sourceConf, err = srcPolicy.Evaluate(ctx)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to evaluate source policy %q: %w", sp.PolicyName(), err)
+		}
+		if err := conf.Merge(cleanConf(sourceConf)); err != nil {
+			return nil, fmt.Errorf("failed to merge config for source policy %q: %w", sp.PolicyName(), err)
+		}
+
+		logsPipelines, err := srcPolicy.LogsPipelines(destPolicy.PreProcessLogIDs(), destPolicy.ExporterIDs(), destPolicy.ExtensionIDs())
```
dashpole, line 239: Source policies construct pipelines passing only `destPolicy.PreProcess<Signal>IDs()` (such as `queuebatch`) and `destPolicy.ExporterIDs()`.

However, the transformation/filter processor (`googlepolicy` or `policy/global`) is not passed into `LogsPipelines`, `MetricsPipelines`, or `TracesPipelines`. As a result, transformation policies are never wired into the pipeline processors list and won't actually execute on telemetry data.

We should ensure the global transformation processor is injected into the pipeline processor sequence (e.g. `receiver -> [pre-processors] -> [policy/global] -> [batch/queue] -> [exporters]`).

pkg/googlepolicy/registry.go
```diff
@@ -0,0 +1,173 @@
...
+	ErrPolicyFailedToLoad          = errors.New("failed to load policy")
+	ErrPolicyFailedValidation      = errors.New("policy validation failed")
+)
+
+// policyRegistry is a central map that tracks all supported policies registered by any components.
+var policyRegistry = map[string]PolicyDriver{}
+
+var (
+	policySetMu        sync.RWMutex
+	activePolicySet    *PolicySet
```
dashpole: Storing `activePolicySet`, `previousPolicySets`, and `watcherChannels` in package-level globals creates mutable shared state across the process.

While having a single active policy set may work for production, this prevents multiple instances from running in parallel in tests (`t.Parallel()`) and can lead to cross-test pollution or race conditions.

Consider encapsulating this state within an instance struct (e.g., on the `provider` or an engine/informer instance) rather than package-level globals so multiple instances can be instantiated and tested independently.

pkg/googlepolicy/policy.go
```diff
@@ -0,0 +1,203 @@
...
+	PreProcessMetricIDs() []component.ID
+	PreProcessLogIDs() []component.ID
+	PreProcessTraceIDs() []component.ID
+	ExtensionIDs() []component.ID
+}
+
+// PolicyDriver is the interface that is used to load a policy
+// object of a known policy type.
+type PolicyDriver interface {
+	LoadPolicy(raw map[string]any) (Policy, error)
```
dashpole, line 81: Our policy schemas are defined as strongly typed Protobufs in `proto/policy/v1alpha1/` (e.g. `LogFilterPolicy`, `MetricFilterPolicy`). Deserializing raw JSON via `mapstructure` / `map[string]any` does not support protobuf generated types (`oneof` matchers, timestamps/durations, field masks).

Instead, we should work with Protobuf messages directly (example interface below):

1. **Update `PolicyDriver` to accept `proto.Message`:**
   ```go
   type PolicyDriver interface {
       TypeURL() string
       Class() PolicyClass
       Validate(policy proto.Message) error
       GenerateConfig(policy proto.Message, ctx *CompilationContext) (*ConfigFragment, error)
   }
   ```
2. **Standardize on `TypedExtensionConfig` (`anypb.Any`):**
   Wrap policies in an Envoy `v3.TypedExtensionConfig` envelope (`type_url` + serialized proto `value`), whether ingested from local files or xDS.
3. **Parse with `protojson` & `proto.Unmarshal`:**
   In the file/xDS ingestor, decode files directly into the protobuf envelope using `protojson.Unmarshal` (for JSON) and `proto.Unmarshal` (for binary).
4. **Driver Unpacking:**
   Look up the driver by `TypeURL()`, unpack `anypb.Any` into the concrete generated message (`anypb.UnmarshalTo`), and call `Validate` / `GenerateConfig`.

pkg/googlepolicy/registry.go
```diff
@@ -0,0 +1,173 @@
...
+
+var (
+	ErrPolicyTypeAlreadyRegistered = errors.New("policy type already registered")
+	ErrPolicyTypeNotFound          = errors.New("no driver found for policy type")
+	ErrPolicyFailedToLoad          = errors.New("failed to load policy")
+	ErrPolicyFailedValidation      = errors.New("policy validation failed")
+)
+
+// policyRegistry is a central map that tracks all supported policies registered by any components.
+var policyRegistry = map[string]PolicyDriver{}
```
dashpole, line 32: Rather than using a package-level global `var policyRegistry = map[string]PolicyDriver{}`:

1. Create a `PolicyDriverRegistry` struct instance.
2. Provide a `NewDefaultRegistry()` factory that registers standard drivers (`GcpDestinationDriver`, `OtlpSourceDriver`, `LogFilterDriver`, etc.).
3. Inject the registry into the `Provider` and `Compiler` instances.

This avoids race conditions during driver registration and enables complete test isolation with `t.Parallel()`, allowing tests to register mock or custom drivers without cross-test interference.

PR #644, opened 2026-09-04
Update Golang version in presubmit.yaml

franciscovalentecastro:
Update Golang version in presubmits.

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -24,7 +24,7 @@ on:
 
 env:
   # Default minimum version of Go to support.
-  DEFAULT_GO_VERSION: 1.25
+  DEFAULT_GO_VERSION: 1.26
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


XuechunHou (approved):

.github/workflows/presubmit.yaml
rafaelwestphal: Can you investigate these failures and  see if they are just noise or not? THanks

rafaelwestphal (changes requested):

.github/workflows/presubmit.yaml
franciscovalentecastro: They are not noise. This is related to the `zizmor` security tool (i'm not fully sure of the whole purpose of the tool), but in summary it fails because a new Google wide policy is that you can't use package like this `actions/checkout@v2`, you have to tag a specific commit with a full SHA (e.g. actions/checkout@fsdafudsgfusf453872uhsaifhiushdfsa ) .

I was doing this in a separate PR. I was only doing to fix a vulnerability, but we don't really need to fix this at the moment (the failure in the other PR was unrelated to the GO versions).

See https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/645

franciscovalentecastro:
I'll convert this PR to a draft. There is a failure in one of the tests which is not very informative. The only thing i'm changing is the Go versions, so it could be related to Go 1.27 still not working nicely with a lot of tools (it's very new).

PR #652, opened 2026-09-08
Cleanup from policy POC PR

braydonk:
* Accidentally left in a distribution yaml file I was using but got rid of
* Change how testing steps work in GitHub Actions to include `pkg/googlepolicy` and any future packages created


rafaelwestphal (approved):

XuechunHou (approved):

.github/workflows/presubmit.yaml
```diff
@@ -55,25 +55,8 @@ jobs:
...
-          fetch-depth: 0
-          lfs: true
-      - uses: actions/setup-go@v5
-        with:
-          go-version: ${{ matrix.go }}
-          cache: true
-      - name: Test Components
-        run: make test-${{matrix.distro}}-components
+      - name: Run tests
+        run: make test-all
```
dashpole, line 59: The presubmit failure is because `target-all-modules` in the root `Makefile` does not exclude `integration_test/gce-testing-internal`, which fails without GCP credentials.

braydonk:
We get this failure **constantly** in Collector Contrib too: https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/actions/runs/34358726841/job/102489944796?pr=652#step:4:37

I don't know what has been going on with the sum database recently but it constantly chokes and flakes like this.

PR #653, opened 2026-09-08
test(gce): enable RetryAlways on storageClient

rafaelwestphal:
### Summary
Configures `storageClient` with `storage.WithPolicy(storage.RetryAlways)` and hardens `gce.UploadContent()` object cleanup so transient Google Cloud Storage (GCS) HTTP 503 (`Service Unavailable`) errors do not abort VM setup across integration test suites.

### Root Cause
Under concurrent CI load, parallel test VMs simultaneously invoke `gce.UploadContent()` to transfer agent configuration files, scripts, and test logs via GCS (`stackdriver-test-143416-file-transfers`). Previously:
1. `storage.NewClient(ctx)` defaulted to `storage.RetryIdempotent`, causing non-idempotent operations like `object.Delete(ctx)` to fail immediately on transient GCS HTTP 503 errors without retrying.
2. Even when file upload and VM transfer succeeded (`err == nil`), any error in `defer object.Delete(ctx)` overwrote `err` and failed the test.

### Changes
1. **`storageClient.SetRetry(storage.WithPolicy(storage.RetryAlways))`**: Enables native retries across all Go `storageClient` operations (`Writer` uploads and `Delete` calls) using `cloud.google.com/go/storage`'s built-in `storage.ShouldRetry` policy.
2. **Harden Object Cleanup**: In `defer object.Delete(ctx)`, ignore `storage.ErrObjectNotExist` and log a warning rather than failing the test if cleanup fails after a successful file transfer (`err == nil`).

### Fixes 
Bug: b/558788032

### Testing
I've tested Ops Agent tests with this change
https://github.com/GoogleCloudPlatform/ops-agent/pull/2425


LujieDuan (approved):
LGTM!

PR #660, opened 2026-09-11
Implement TransformationPolicy interface and native googlepolicyprocessor

rafaelwestphal:
### Description
This PR introduces the base transformation policy architecture and the `googlepolicyprocessor` component for the Google-built OpenTelemetry Collector (GBOC).

### Summary of Changes
1. **`pkg/googlepolicy`**:
   - Introduces the `TransformationPolicy` interface supporting `TransformLogs`, `TransformMetrics`, and `TransformTraces`.
   - Adds `PolicySet.TransformationPolicies()` helper to retrieve all transformation policies from a policy set.
2. **`googlepolicyprocessor`**:
   - Implements native in-place transformers for all three telemetry signals (`log_transformer.go`, `metric_transformer.go`, `trace_transformer.go`) without allocating intermediate slices.
   - Introduces an atomic, thread-safe evaluator (`evaluator.go`) that watches for policy updates via `googlepolicy.RegisterWatcherChannel` and compiles new rule sets dynamically.
   - Wires the evaluator into the processor lifecycle (`ConsumeLogs`, `ConsumeMetrics`, `ConsumeTraces`).
   - Comprehensive unit test coverage for evaluator compilation, concurrent policy reloading, signal transformations, and processor lifecycle.

### Testing
- All unit tests passing under `pkg/googlepolicy/...` and `components/.../googlepolicyprocessor/...`.
- `go fmt` applied across all modified packages.


XuechunHou:
# PR #660 Review: TransformationPolicy & GooglePolicyProcessor

**PR Link**: [GoogleCloudPlatform/opentelemetry-operations-collector#660](https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pull/660)  
**Author**: @rafaelwestphal  
**Reviewed by**: @XuechunHou  

---

## Executive Summary & Architecture Evaluation

This PR introduces the base transformation policy architecture and the `googlepolicyprocessor` component for the Google-Built OpenTelemetry Collector (GBOC). It enables dynamic, in-flight compilation and evaluation of declarative Google telemetry policies (`LogFilterPolicy`, `MetricFilterPolicy`, `TraceFilterPolicy`) without requiring collector restarts.

### Architectural Strengths
1. **Lock-Free Hot Path with Atomic Evaluator Swapping**:
   * Compiling rules and regexes is performed asynchronously in a watcher goroutine upon policy set updates.
   * Telemetry processing pipelines access the active engine via `atomic.Pointer[Evaluator]`, ensuring the hot data path remains completely non-blocking and lock-free.
2. **In-Place Mutation & Pruning (`RemoveIf`)**:
   * Telemetry batches are evaluated directly within OTel's native `RemoveIf()` across Resource -> Scope -> Record/DataPoint/Span hierarchies. This avoids intermediate slice allocations or deep-copy overhead for passing records.
3. **Declarative Protobuf Contract**:
   * Decouples policy definition from ingestion transport (xDS, file watcher, or control plane), establishing a clean boundary between management plane schemas and collector runtime processing.

### Key Architectural Recommendations
1. **Segregate Instrument-Level from DataPoint-Level Metric Policies**:
   * *Current Design*: Evaluates all metric policies inside `datapoints.RemoveIf`.
   * *Why It’s a Problem*:
     * **Performance**: A policy matching on metric name (e.g. `^drop_jvm_.*`) evaluates the regex on every single data point (e.g. 10,000 times for a large histogram or gauge batch) instead of once per metric instrument.
     * **Correctness**: If a metric has 0 data points, the data point loop never executes, causing empty metrics to be dropped unconditionally—even when no policy matched or when an `ACTION_KEEP` policy matched.
   * *Recommendation*: Split metric evaluation into an instrument-level pass (evaluated once on `pmetric.Metric`) and a data-point-level pass (evaluated only when matchers target `datapoint_attribute`).
2. **Deterministic Policy Ordering**:
   * `TransformationPolicies()` iterates over Go maps with randomized order. Sorting policies deterministically (e.g. by `PolicyName()` or ID) ensures predictable compilation across reloads and future-proofs the engine for ordered stateful mutations.
3. **Extensibility of `TransformationPolicy`**:
   * `TransformationPolicy` in `pkg/googlepolicy` is generic, but `NewEvaluator` uses a concrete type switch for the three v1alpha1 filter protos. Formalizing compilation hooks will make adding future transformation policies (e.g. masking, sampling) cleaner without central switch explosion.

---

## High-Priority / Blocking Issues

### 1. Data Loss: Empty Metrics (0 Datapoints) Unconditionally Dropped
In `metric_transformer.go` (`transformMetricDataPoints`):
```go
case pmetric.MetricTypeGauge:
    e.transformNumberDataPoints(m, m.Gauge().DataPoints(), ...)
    return m.Gauge().DataPoints().Len() == 0
```
In `TransformMetrics`, `sm.Metrics().RemoveIf(...)` removes any metric where `transformMetricDataPoints` returns `true`.
* **Issue**: If an incoming metric stream contains 0 data points (e.g. an empty gauge or sum before scrape/initialization), `DataPoints().Len() == 0` evaluates to `true`.
* **Impact**: The processor immediately drops the metric from the stream **even if no drop policy matched**, and **even if an `ACTION_KEEP` policy explicitly matched it**. Per the proto specification, metrics should only be dropped if an `ACTION_DROP` policy actively pruned all of its points.

### 2. Policies with No Matchers Silently Drop All Telemetry
In `evaluator.go`, `compileLogPolicy`, `compileMetricPolicy`, and `compileTracePolicy` do not validate `len(p.GetMatches()) > 0`.
* **Issue**: `compiledLogPolicy.matches()` iterates over `p.matchers`. If `matchers` is empty, it returns `true` for every single record.
* **Impact**: An `ACTION_DROP` policy with an empty `matches` list will drop 100% of telemetry in the collector pipeline. Per `log_filter_policy.proto`, a policy with no matchers is invalid.
* **Fix**: Validate that `len(p.GetMatches()) > 0` and `p.GetAction() != Action_ACTION_UNSPECIFIED` at policy compile time.

---

## Schema & Specification Non-Conformance

### 3. Casing Mismatch in Metric Type and Temporality
`proto/policy/v1alpha1/metric_filter_policy.proto` specifies:
* `METRIC_DESCRIPTOR_FIELD_TYPE`: string predicates evaluate against **uppercase** type names (`"GAUGE"`, `"SUM"`, `"HISTOGRAM"`, `"EXPONENTIAL_HISTOGRAM"`, `"SUMMARY"`).
* `METRIC_DESCRIPTOR_FIELD_AGGREGATION_TEMPORALITY`: string predicates evaluate against **uppercase** names (`"DELTA"`, `"CUMULATIVE"`).

In `metric_transformer.go`:
`metricTypeToString` and `temporalityToString` return lowercase strings (`"gauge"`, `"sum"`, `"delta"`, `"cumulative"`), unlike `trace_transformer.go` which correctly returns uppercase (`"INTERNAL"`, `"SERVER"`).
* **Impact**: Valid policies created according to the protobuf specification will fail to match.

### 4. `METRIC_DESCRIPTOR_FIELD_IS_MONOTONIC` is Unimplemented
`MetricDescriptorField_METRIC_DESCRIPTOR_FIELD_IS_MONOTONIC` (field #6 in `metric_filter_policy.proto`) is omitted from the switch statement in `extractMetricTarget` in `metric_transformer.go`, falling into `default: return nil, false`.

### 5. Array / Slice Indexing in `AttributePath` Fails
In `common_policy_types.proto`:
> *"When traversing array/list values, string segments representing non-negative integers (e.g. "0", "100") are interpreted as zero-based array indices (e.g. ["items", "0", "id"])."*

In `evaluator.go` (`lookupPath`):
```go
if current.Type() != pcommon.ValueTypeMap {
    return nil, false
}
```
Traversing into a slice (`pcommon.ValueTypeSlice`) aborts immediately. Slices with integer index segments are not supported.

### 6. Root Span `parent_span_id` String Equality Matching
In `trace_filter_policy.proto`:
> *"When evaluated with `exists`, this evaluates to false for root spans... When evaluated against string predicates (`exact`, `regex`), a root span matches against the empty string `""`."*

In `trace_transformer.go`, `extractTraceTarget` returns `nil, false` when `psid.IsEmpty()`. In `evaluator.go`, string predicates require `exists == true`. As a result, checking `equals: ""` or `regex: "^$"` against a root span's `parent_span_id` always fails.

---

## Semantics & Robustness

### 7. Coupling Field Presence (`exists`) with Empty Strings
In `extractLogTarget`, `extractMetricTarget`, and `extractTraceTarget`, first-class string fields (`SeverityText`, `Description`, `StatusMessage`) return `exists = (s != "")`.
Because `predEquals` requires `exists == true`, matching `equals: ""` will always evaluate to `false`. The schema defines that absent/empty evaluates to false *specifically for the `exists` predicate*, not that empty fields are excluded from equality checks.

### 8. Loss of Log Record Body Typing via `AsString()`
In `log_transformer.go`:
```go
case policyv1alpha1.LogRecordField_LOG_RECORD_FIELD_BODY:
    s := ctx.Record.Body().AsString()
    return s, s != ""
```
OTel log bodies can be structured types (`int64`, `bool`, `bytes`, `slice`, `map`). Converting unconditionally to a string breaks typed matching (e.g. `Value_IntValue`, `Value_BytesValue` in `matchEquals`, or binary `matchContains`). `pcommonValueToAny` should be used instead.

### 9. Regex on Numeric and Boolean Fields
`predRegex` in `evaluator.go` only inspects `val.(string)`. For numeric fields (`SEVERITY_NUMBER`) or boolean fields (`IS_MONOTONIC`), `val` is `int64` or `bool`. The schema explicitly specifies that regex on severity number evaluates against canonical base-10 strings, and booleans against `"true"`/`"false"`.

### 10. Gauge & Summary Temporality Reports `exists = true`
For Gauges and Summaries, `AggregationTemporality` is unspecified. Currently, `extractMetricTarget` returns `"unspecified", true`. An `exists` check on temporality evaluates to `true` for Gauges and Summaries, even though temporality does not exist on those metric types.

### 11. Non-Deterministic Compilation Order from Map Iteration
`TransformationPolicies()` in `pkg/googlepolicy/policy.go` iterates over `ps.Policies` (`map[string]*PolicySetEntry`). Because Go map iteration order is randomized, policy ordering changes across reloads. Sorting by `PolicyName()` or policy ID provides deterministic behavior.

### 12. Lifecycle: Redundant `shutdown` Causes Panic
In `processor.go`, `close(p.stopCh)` is called directly without a `sync.Once`. If `shutdown()` is called multiple times by the collector or in unit tests, it will panic with `close of closed channel`.

---

## Performance Considerations

1. **Instrument-Level vs. DataPoint-Level Policies**: Metric policies matching only descriptor fields (e.g. metric name) are currently evaluated on every single data point in `datapoints.RemoveIf`. For instance, a policy matching `^drop_jvm_.*` evaluates the regex on every single data point (e.g. 10,000 times for a large histogram or gauge batch) rather than once at the descriptor level. Segregating instrument-level policies allows dropping matching metric streams in $O(1)$ without iterating millions of data points.
2. **Short-Circuit on `ACTION_KEEP`**: In `EvalLog`, `EvalMetric`, and `EvalTrace`, evaluating subsequent policies after an `ACTION_KEEP` match is unnecessary since `"KEEP always overrides DROP"`. Short-circuiting saves CPU in the hot path.
3. **Heap Allocations in Hot Path**: Boxing every field into `any` and allocating `[]any` in `pcommonValueToAny` on every record creates heap churn. Operating directly on `pcommon.Value` avoids these allocations.

---

## Test Coverage Gaps

`evaluator_test.go` currently only tests string equality and one regex. Recommended additions:
* Predicates: `exists`, `gt`, `gte`, `lt`, `lte`, `contains`
* Targets: `ResourceAttribute`, `ScopeAttribute`, array index traversal in `AttributePath`
* First-class fields: `SeverityNumber`, `TraceID`, `SpanID`, `ParentSpanID` (root spans), `ScopeField`
* Multi-policy evaluation: Verifying `ACTION_KEEP` overrides `ACTION_DROP`
* Compilation errors: Malformed regex, missing matchers, unspecified actions

XuechunHou (approved):

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/evaluator.go
```diff
@@ -0,0 +1,405 @@
...
+		if err != nil {
+			return pred, fmt.Errorf("invalid regex '%s': %w", p.Regex, err)
+		}
+		pred.typ = predRegex
+		pred.regex = re
+	case *policyv1alpha1.LogMatcher_Gt:
+		pred.typ = predGt
+		pred.gtVal = p.Gt
+	case *policyv1alpha1.MetricMatcher_Gt:
+		pred.typ = predGt
```
dashpole, line 147: Returning `nil, err` on the first compile error disables all other valid policies in the PolicySet (leaving `p.evaluator` as `nil` on startup), whereas OTEP 4738 and the design doc specify that invalid policies should be skipped while remaining valid policies continue to execute. Consider collecting errors into a slice and returning `ev, errors.Join(errs...)` so `reloadPolicies` can log the error while still activating the valid policies.

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/log_transformer.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, errors.New("policy action must be specified")
+	}
+	if len(p.GetMatches()) == 0 {
+		return nil, errors.New("policy must have at least one matcher")
+	}
+	cp := &compiledLogPolicy{
+		id:     p.GetId(),
+		action: p.GetAction(),
+	}
+	for _, m := range p.GetMatches() {
```
dashpole, line 47: If a matcher has `m.GetTarget() == nil` or `m.GetTarget().GetTarget() == nil`, compilation succeeds and `extractLogTarget` returns `nil, false` at runtime (which matches every record if `negate: true` is set). Consider validating that `m.GetTarget()` and its inner `Target` oneof are non-nil here (and in `compileMetricPolicy` and `compileTracePolicy`).

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/log_transformer.go
```diff
@@ -0,0 +1,194 @@
...
+			val := pcommonValueToAny(ctx.Record.Body())
+			if s, ok := val.(string); ok {
+				return s, s != ""
+			}
+			return val, true
+		case policyv1alpha1.LogRecordField_LOG_RECORD_FIELD_SEVERITY_TEXT:
+			s := ctx.Record.SeverityText()
+			return s, s != ""
+		case policyv1alpha1.LogRecordField_LOG_RECORD_FIELD_SEVERITY_NUMBER:
+			n := int64(ctx.Record.SeverityNumber())
```
dashpole, line 143: When `SeverityNumber()` is unset (`0`), returning `int64(0), false` causes `evaluate` (`hasVal := exists || val != nil`) to treat unset severities as numeric `0`, meaning a rule like `lt: 9` (drop debug/trace logs) will accidentally drop all logs with unspecified severity. Returning `nil, false` when unspecified avoids matching numeric predicates:
```suggestion
		case policyv1alpha1.LogRecordField_LOG_RECORD_FIELD_SEVERITY_NUMBER:
			if ctx.Record.SeverityNumber() == plog.SeverityNumberUnspecified {
				return nil, false
			}
			return int64(ctx.Record.SeverityNumber()), true
```

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/metric_transformer.go
```diff
@@ -0,0 +1,349 @@
...
+		Scope:                  scope,
+		ResourceSchemaURL:      resourceSchemaURL,
+		ScopeSchemaURL:         scopeSchemaURL,
+	}
+
+	if initialLen == 0 || !e.hasDatapointMetricPolicies {
+		return e.evalMetricInstrumentOnly(instrumentCtx)
+	}
+
+	for _, p := range e.metricPolicies {
```
dashpole, line 127: When `hasDatapointMetricPolicies` is true, instrument-level `ACTION_DROP` policies are not short-circuited here, and `EvalMetric` below re-evaluates all `!p.isDatapointLevel` policies (including regexes on metric name or resource attributes) on every single datapoint. Consider evaluating `!p.isDatapointLevel` policies once on `instrumentCtx` to get an initial `instrumentDrop` bool, short-circuiting if `instrumentDrop` is true and no datapoint-level `ACTION_KEEP` policy exists, and only evaluating `p.isDatapointLevel` policies inside the datapoint loop.

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/trace_transformer.go
```diff
@@ -0,0 +1,223 @@
...
+				return "", false
+			}
+			return psid.String(), true
+		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_STATUS_MESSAGE:
+			s := ctx.Span.Status().Message()
+			return s, s != ""
+		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_KIND:
+			return spanKindToString(ctx.Span.Kind()), true
+		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_STATUS_CODE:
+			return spanStatusToString(ctx.Span.Status().Code()), true
```
dashpole, line 157: Per `trace_filter_policy.proto`, `exists` on `status_code` should evaluate to `false` when the status is `UNSET`. Returning `c != ptrace.StatusCodeUnset` fixes `exists` while still allowing `equals` or `regex` to match `"UNSET"`:
```suggestion
		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_STATUS_CODE:
			c := ctx.Span.Status().Code()
			return spanStatusToString(c), c != ptrace.StatusCodeUnset
```

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/processor_test.go
```diff
@@ -58,3 +62,292 @@ func TestProcessLogs(t *testing.T) {
...
+	require.Equal(t, 1, out.ResourceSpans().Len())
+	require.Equal(t, 1, out.ResourceSpans().At(0).ScopeSpans().Len())
+	spans := out.ResourceSpans().At(0).ScopeSpans().At(0).Spans()
+	require.Equal(t, 1, spans.Len())
+	assert.Equal(t, "/api/v1/checkout", spans.At(0).Name())
+}
+
+func TestProcessLogs_DynamicPolicyUpdate(t *testing.T) {
+	// Start processor with empty policy set
+	googlepolicy.SetActivePolicySet(nil)
```
dashpole, line 247: `googlepolicy.SetActivePolicySet(nil)` is a no-op in `pkg/googlepolicy/registry.go` (`if policySet == nil { return }`), so `activePolicySet` leaks across tests and causes `go test -count=2 .` to fail on the second run of this test. Consider adding a `ResetActivePolicySet()` helper (or allowing `SetActivePolicySet` to clear state or setting an empty `&googlepolicy.PolicySet{}`) so tests clean up properly.

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/processor.go
```diff
@@ -35,17 +44,88 @@ func newGooglePolicyProcessor(cfg *Config, logger *zap.Logger) *googlePolicyProc
...
+func (p *googlePolicyProcessor) reloadPolicies() {
+	ps := googlepolicy.ActivePolicySet()
+	if ps == nil {
+		p.evaluator.Store(nil)
+		return
+	}
+
+	policies := ps.TransformationPolicies()
+	ev, err := NewEvaluator(policies)
+	if err != nil {
```
dashpole, line 70: Paired with `NewEvaluator` collecting compile errors rather than aborting: if `err != nil` but `ev != nil`, returning early here leaves `p.evaluator` as `nil` on startup (disabling all valid policies in the set). Consider logging `err` when non-nil and still calling `p.evaluator.Store(ev)` whenever `ev != nil` so valid policies remain enforced.

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/trace_transformer.go
```diff
@@ -0,0 +1,223 @@
...
+		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_NAME:
+			s := ctx.Span.Name()
+			return s, s != ""
+		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_TRACE_ID:
+			tid := ctx.Span.TraceID()
+			if tid.IsEmpty() {
+				return nil, false
+			}
+			return tid.String(), true
+		case policyv1alpha1.SpanRecordField_SPAN_RECORD_FIELD_SPAN_ID:
```
dashpole, line 139: Non-urgent future optimization idea: calling `tid.String()` (and `sid.String()` here and in `log_transformer.go`) allocates a new hex string on the heap for every evaluated record. For exact matching at high throughput, encoding onto a stack-allocated `[32]byte` or `[16]byte` buffer and comparing bytes directly could help reduce GC pressure down the road.

dashpole (commented):
For future consideration

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/log_transformer.go
rafaelwestphal: Fixed in #667! LOG_RECORD_FIELD_SEVERITY_NUMBER now returns (nil, false) when ctx.Record.SeverityNumber() == plog.SeverityNumberUnspecified, ensuring numeric range predicates like lt: {int_value: 9} evaluate to EvalNoMatch on logs without a severity number while still dropping SeverityNumberDebug (5). Added TestUnsetSeverityNumberReportsAbsent to pin this behavior.

PR #661, opened 2026-09-13
fix(googlecontrolplaneprovider): build built-in policy configs from factory defaults

XuechunHou:
## Summary

The built-in destination and self-metrics policies constructed their component
configs as struct literals. Every field the corresponding factory would have
populated was therefore silently zeroed, and the generated collector config
could not start.

Four separate failures came from this one habit:

| # | Symptom | Cause |
|---|---------|-------|
| 1 | `unknown type: "queue_batch"` — config fails to unmarshal | processor type hardcoded as `"queue_batch"`, but the component registers as `"queuebatch"` |
| 2 | `failed to create logger: no encoder name specified` | `otelconftelemetry.Config` literal dropped `Logs.Encoding` |
| 3 | `authenticator not found` | `googleclientauth` declared under `extensions` but never listed in `service::extensions` |
| 4 | exporter to `telemetry.googleapis.com:443` had retries disabled, no sending queue, no gzip, 0s timeout | `otlpexporter.Config` literal |

Bugs 1–3 prevented startup outright. Bug 4 did not — it would just have quietly
dropped telemetry on any transient failure.

In every case the fix is the same: start from `CreateDefaultConfig()` and
override only the fields the policy actually cares about. Fixing #2 also
recovered log sampling and the stderr output paths, which the literal had
discarded without any symptom.

For #1 the type now comes from `queuebatchprocessor.NewFactory().Type()` rather
than a string, so it cannot drift from the component again.

## On the golden file

`testdata/policysets/builtin/result.yaml` had frozen the broken output, so the
test passed while the product could not start. It is regenerated here.

Worth calling out: a golden comparison **cannot** catch this class of bug on its
own, because it never asks the collector to load the config it just produced. It
only checks that the YAML matches what the YAML used to be. A follow-up test
that feeds the generated config to `otelcol` and asserts it loads would close
this gap; I have not added one here to keep the diff focused.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` — all clean
- `go test ./...` — passes
- Full repo pre-commit hook (lint, license, misspell, full distro build) — passes
- Manually ran the collector against the built-in policy set. It now starts
  clean with no config workarounds, and the OTLP exporter connects to
  `telemetry.googleapis.com:443`.

## Scope

Three files, no dependency changes. Unrelated in-flight work on the xDS policy
manager is deliberately excluded.


LujieDuan (approved):
LGTM

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/selfmetrics/selfmetrics_policy.go
```diff
@@ -89,61 +89,69 @@ func (p *SelfMetricsPolicy) Evaluate(ctx context.Context) (*confmap.Conf, error)
...
+		},
+	}
+
+	telemetryCfg.Logs.Level = zapcore.InfoLevel
+	telemetryCfg.Logs.Processors = []config.LogRecordProcessor{
+		{
+			Batch: &config.BatchLogRecordProcessor{
+				Exporter: config.LogRecordExporter{
+					OTLP: &config.OTLP{
+						Endpoint: &endpoint,
```
dashpole, line 121: When `p.Port` is `0` (as in `BuiltInSelfMetricsPolicy`), `otlpreceiver` binds to an OS-assigned ephemeral port on `localhost:0`, while this OTLP exporter literally dials TCP port `0` (`http://localhost:0`) and fails with `connection refused`. Consider defaulting `p.Port` to a fixed loopback port or reserving an available port when `p.Port == 0` so self-telemetry reaches the receiver.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/gcpdestination/gcpdestination_policy.go
```diff
@@ -75,22 +76,47 @@ func (p *GCPDestinationPolicy) Evaluate(_ context.Context) (*confmap.Conf, error
...
+	// buffer in front of the same export path.
+	//
+	// Setting None here is only half the job: a None Optional marshals to nil,
+	// the nil key is then dropped from the generated config, and a config that
+	// simply omits `sending_queue` gets the factory default (queue enabled)
+	// back when the collector loads it. The disable has to be written out
+	// explicitly, which is done after marshaling below.
+	otlpExporter.QueueConfig = configoptional.None[exporterhelper.QueueBatchConfig]()
+
 	otlpExporterType, _ := component.NewType("otlp_grpc")
```
dashpole, line 108: Quick nit: since `otlpexporter.NewFactory()` is already used above, we can grab the component type from the factory here too (and similarly for `googleclientauthextension.NewFactory().Type()` on line 72) to avoid hardcoded type strings:
```suggestion
	otlpExporterType := otlpexporter.NewFactory().Type()
```

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/gcpdestination/gcpdestination_policy.go
```diff
@@ -75,22 +76,47 @@ func (p *GCPDestinationPolicy) Evaluate(_ context.Context) (*confmap.Conf, error
...
 	}
 
-	queueBatchType, _ := component.NewType("queue_batch")
+	// Taken from the factory rather than spelled out, so the generated config
+	// cannot drift from the type the processor actually registers under. A
+	// hardcoded "queue_batch" here silently produced a config that failed to
+	// unmarshal once the component settled on "queuebatch".
+	queueBatchType := queuebatchprocessor.NewFactory().Type()
 
 	queueBatchLog := queuebatchprocessor.NewFactory().CreateDefaultConfig().(*queuebatchprocessor.Config)
```
dashpole, line 121: Since this replaces the exporter `sending_queue` (which defaults to 10 consumers and `queue_size: 1000`), note that `queuebatchprocessor` defaults to `num_consumers: 1` and `queue_size: 10` with `block_on_overflow: true`, which could bottleneck export concurrency.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/testdata/policysets/builtin/result.yaml
```diff
@@ -2,16 +2,20 @@ exporters:
     otlp_grpc/default_gcp_destination:
         auth:
             authenticator: googleclientauth/default_gcp_destination
-        balancer_name: ""
+        balancer_name: round_robin
```
dashpole, line 5: Since all five component factories used by the built-in policies are already direct dependencies in `go.mod`, we could catch config unmarshaling regressions directly in `TestPolicySets` by passing the retrieved `confmap` to `otelcol.NewConfigProvider` and calling `cfg.Validate()`.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/gcpdestination/gcpdestination_policy.go
```diff
@@ -75,22 +76,47 @@ func (p *GCPDestinationPolicy) Evaluate(_ context.Context) (*confmap.Conf, error
...
-			Auth: configoptional.Some(configauth.Config{
-				AuthenticatorID: authID,
-			}),
-		},
-	}
+	// Declaring the extension above only defines it; the collector instantiates
+	// nothing that is not also listed under service::extensions. Without this
+	// the exporter's authenticator reference below cannot be resolved and
+	// startup fails with "authenticator not found".
+	conf.Service.Extensions = extensions.Config{authID}
```
dashpole, line 83: Thought on the overall config assembly architecture: right now each policy builds a mini `otelcol.Config`, calls `cm.Marshal(conf)`, and `provider.go` merges those maps via `conf.Merge()`. That pattern introduces a few structural issues:
1. Slice clobbering on merge: `confmap.Conf.Merge()` overwrites slices instead of appending them, so setting `conf.Service.Extensions` here means any subsequent policy that declares an extension will clobber earlier extensions.
2. Marshal/unmarshal asymmetry: `confmap.Marshal` omits `configoptional.None` fields, which causes `otelcol` to re-apply factory defaults (like `sending_queue.enabled = true`) on unmarshal and requires manual post-marshal map patches like `disableSendingQueue`.
3. Missing component validation: `confmap.Validate(conf)` in `provider.go` only checks map structure and never calls component `.Validate()` methods or verifies pipeline/extension cross-references.

Goals:
- Keep compile-time type safety and factory defaults (`CreateDefaultConfig()`) when building component configs in policy drivers.
- Ensure correct, non-destructive merge behavior for slice/set fields (`service::extensions` and pipeline processor chains).
- Run real component `.Validate()` and full topology cross-reference validation before returning from `Retrieve()`.

Proposed design overview (Typed Fragments + Central Assembler):
- Instead of returning a marshaled `*confmap.Conf`, each policy's `Evaluate()` returns a typed `PolicyFragment` struct containing its typed component maps (`Receivers`, `Processors`, `Exporters`, `Extensions`), a `ServiceExts []component.ID` slice, and pipeline definitions, calling `component.ValidateConfig` on each component struct immediately.
- A central Assembler in `provider.go` merges all `PolicyFragment` structs into a single unified `*otelcol.Config` in memory—appending and deduplicating `ServiceExts` and ordering pipeline processors cleanly—before calling `confmap.Marshal` once at the very end.
- A generic helper (comparing `configoptional.Optional` fields against `factory.CreateDefaultConfig()`) can automatically emit `enabled: false` for disabled optionals, followed by a final validation gate (`otelcol.ConfigProvider` / `cfg.Validate()`) on the assembled config.

dashpole (commented):
Comments for future consideration

PR #662, opened 2026-09-14
feat(googlepolicy): add xDS policy manager for control-plane-driven config

XuechunHou:
## What

Adds an xDS-based policy manager so the collector can receive its policy set from a control plane over an ADS stream, alongside the existing file-based manager.

`pkg/googlepolicy/xds_policy_manager.go` implements the `googlepolicy.Manager` interface on top of an Aggregated Discovery Service stream:

- **Subscribe & decode** — subscribes to `google.telemetry.xds.v1alpha1.TelemetryCollector` keyed by the collector and fleet identity, unwraps each `TypedExtensionConfig` into the generic raw policy map the existing policy drivers already consume, and installs the result as the active policy set.
- **ACK/NACK** — a policy this collector cannot load is NACKed back to the control plane rather than silently ignored, so bad policy is visible where it was authored.
- **Auth** — authenticates with ADC, falling back to an ID token. Transport security can be disabled for local testing with `?insecure=true`.
- **Reconnect** — jittered backoff on stream failure, and the confmap watcher is notified on every accepted update so the collector reloads.

The provider gains an `xds` inner scheme:

```
googlecontrolplane:xds://control-plane.example.com:443
```

## Fail-open behaviour

If the control plane is unreachable, the retry loop runs in the background and `Retrieve` still returns, so the collector starts on its built-in policies instead of failing startup. Covered by `TestRetrieve_XDSServerUnreachable`, which points the provider at a dead port and asserts the built-in destination/self-metrics pipelines are still stood up.

## Also

`resolveFleetID` factors out the context-then-env fleet lookup so the new scheme handler and `evaluateActivePolicySet` resolve the fleet identically.

## Testing

- `go test ./...` in `pkg/googlepolicy` and in the `googlecontrolplaneprovider` module — all pass.
- Full `make` pre-commit path (license, lint, `ocb` generate, collector build) passes.

## Notes for reviewers

This is carved out of a larger branch. Deliberately **not** included here:

- The `logfilter` policy driver, which depends on `TransformationPolicy` / `Signal` from #660 — it'll follow once that lands.
- The built-in policy factory-default fixes, which are already up as #661.


pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,832 @@
...
+	// EnforcementPolicy{MinTime: 30s} so these values are honored.
+	keepaliveTime    = 60 * time.Second
+	keepaliveTimeout = 20 * time.Second
+)
+
+var (
+	// Configuration errors, returned by NewXDSPolicyManager.
+	ErrXDSMissingServerAddr   = errors.New("xDS server address cannot be empty in URI")
+	ErrXDSMissingCollectorID  = errors.New("a collector ID is required for the xDS policy manager")
+	ErrXDSMissingFleetID      = errors.New("a fleet ID is required for the xDS policy manager: set the 'fleet' URI query parameter or the FLEET_ID environment variable")
```
dashpole: The error message mentions setting the `fleet` query parameter, but `FleetIDFromURI` only accepts `gcp.fleet_id` and ignores `fleet` (also worth updating the doc comments on lines 168 and 173).
```suggestion
	ErrXDSMissingFleetID      = errors.New("a fleet ID is required for the xDS policy manager: set the 'gcp.fleet_id' URI query parameter or the FLEET_ID environment variable")
```

pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,832 @@
...
+// run maintains the ADS stream for the lifetime of the manager, reconnecting
+// with exponential backoff whenever it drops.
+func (m *xdsPolicyManager) run(ctx context.Context) {
+	defer m.wg.Done()
+
+	node := &corev3.Node{
+		Id:      m.collectorID,
+		Cluster: m.fleetID,
+		Locality: &corev3.Locality{
+			Region: defaultRegion,
```
dashpole: Hardcoding `asia-east1` means every collector reports that region in its xDS `Node.Locality` regardless of where it actually runs, which can misroute locality-aware control plane traffic. Also, per the GBOC policy ingestion design doc, we should populate `node.Extensions` here with the registered policy types so the control plane knows what schemas this collector supports.

pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,832 @@
...
+	if version := resp.GetVersionInfo(); version != "" && version == m.LastAppliedVersion() {
+		m.logger.Debug("Re-ACKing an already applied xDS revision", zap.String("version", version))
+		if err := m.sendACK(stream, node, version, resp.GetNonce()); err != nil {
+			m.logger.Error("Failed to send xDS ACK", zap.String("version", version), zap.Error(err))
+		}
+		return
+	}
+
+	rawPolicies, err := extractRawPolicies(resp)
+	if err != nil {
```
dashpole: Returning early before `SetActivePolicySet` whenever `extractRawPolicies` or `MakePolicySet` returns an error makes policy loading fail-closed (dropping all valid policies in the set if a single policy has an unrecognized type URL). To align with the fail-open requirement in the design doc and OTEP 4738, consider still applying the valid subset in `policySet` even when sending a NACK.

pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,832 @@
...
+//     falls back to ADC with OpenID scopes and extracts the ID token via GoogleIDTokenSource.
+func ResolveTokenSource(ctx context.Context, serverAddr string) (oauth2.TokenSource, error) {
+	host, _, err := net.SplitHostPort(serverAddr)
+	if err != nil {
+		host = serverAddr
+	}
+	audience := "https://" + host
+
+	// 1. GCE / GKE / Service Accounts: metadata server or SA key
+	if idTS, err := idtoken.NewTokenSource(ctx, audience); err == nil {
```
dashpole: On GCE and GKE, `idtoken.NewTokenSource` always succeeds against the metadata server and returns an OIDC ID token, whereas standard Google Cloud gRPC APIs like `telemetrydirector.googleapis.com` expect an OAuth2 access token with `cloud-platform` scope.

dashpole, line 934: Please add a comment or TODO noting that OIDC ID tokens are currently used for the Cloud Run test server and will need to be updated to OAuth2 access tokens with `cloud-platform` scope to work with `telemetrydirector.googleapis.com` or Traffic Director.

pkg/googlepolicy/xds_policy_manager.go
dashpole: Returning early before `SetActivePolicySet` whenever `extractRawPolicies` or `MakePolicySet` returns an error makes policy loading fail-closed (dropping all valid policies in the set if a single policy has an unrecognized type URL). Please make sure this is fail-open per the design doc and OTEP 4738 by still applying the valid policies in `policySet` via `SetActivePolicySet` even when sending a NACK.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
LujieDuan: Can update to resolve project id the same way

pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,1005 @@
...
+	// telling us to drop everything. That is legitimate, but it disables all
+	// policy enforcement, so it is never done quietly.
+	if active := ActivePolicySet(); len(policySet.Policies) == 0 && active != nil && len(active.Policies) > 0 {
+		m.logger.Warn("xDS revision contains no policies, clearing the active policy set",
+			zap.String("version", resp.GetVersionInfo()),
+		)
+	}
+
+	// 2. Activate in memory (googlepolicyprocessor immediately picks this up).
+	SetActivePolicySet(policySet)
```
dashpole: When `extractErr` or `makeErr` is non-nil on a partially valid revision, we should still send a NACK after applying `policySet` via `SetActivePolicySet` so the control plane receives the `ErrorDetail` for the rejected policies instead of an ACK:
```suggestion
	SetActivePolicySet(policySet)
	if err := errors.Join(extractErr, makeErr); err != nil {
		m.sendNACK(stream, node, resp.GetNonce(), err)
		m.markReady()
		return
	}
```

pkg/googlepolicy/policy.go
```diff
@@ -172,6 +173,28 @@ type PolicyDriver interface {
...
+// policy arrives as a bare google.protobuf.Any whose body carries no "type"
+// field -- LogFilterPolicy and friends have no such field -- so the message
+// identity in the type URL is the only routing information available. Every
+// other delivery path has the policy type written out by whoever authored the
+// config, which is the assumption MakePolicySet documents.
+//
+// Drivers whose policy has no proto representation, such as the built-in
+// source and destination policies that are plain Go structs, simply do not
+// implement this and remain unreachable over xDS.
+type ProtoPolicyDriver interface {
```
rafaelwestphal, line 192: Since `rawFromAny` (`xds_policy_manager.go:838`) already unmarshals the binary `anypb.Any` into the concrete `proto.Message` via `msgAny.UnmarshalNew()`, could we add `LoadPolicyFromProto(msg proto.Message) (Policy, error)` here and provide a generic `ProtoDriver[P proto.Message]`?
That eliminates the 4x serialization round-trip (`proto` $\rightarrow$ JSON $\rightarrow$ `map[string]any` $\rightarrow$ JSON $\rightarrow$ `proto`), avoids collisions if a future policy proto defines a field named `type`, and lets `logfilter`/`metricfilter`/`tracefilter` delete their duplicated `Driver` structs.

pkg/googlepolicy/registry.go
```diff
@@ -19,18 +19,50 @@ import (
...
 
 // policyRegistry is a central map that tracks all supported policies registered by any components.
 var policyRegistry = map[string]PolicyDriver{}
 
+// policyProtoRegistry routes a policy proto's fully qualified message name to
+// the policy type registered for it. Only drivers implementing
+// ProtoPolicyDriver appear here, which is deliberate: it is the set of policies
+// that can be delivered as a bare proto rather than as an authored config with
+// an explicit "type" field.
+var policyProtoRegistry = map[protoreflect.FullName]string{}
```
rafaelwestphal, line 50: `policySetMu` guards `activePolicySet` and `previousPolicySets`, but `policyRegistry` and `policyProtoRegistry` are read from background xDS stream goroutines (`PolicyTypeForProto` and `LoadPolicy`) without a mutex. Worth guarding both maps with a `sync.RWMutex`.

pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,1005 @@
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.ready != nil && !m.readyClosed {
+		m.readyClosed = true
+		close(m.ready)
+	}
+}
+
+// Stop terminates the background stream loop and waits for it to exit. It is
```
rafaelwestphal, line 434: `Stop()` sets `m.cancel = nil` and unlocks `m.mu` before `m.wg.Wait()` finishes — a concurrent `Start()` can observe `m.cancel == nil` and call `m.wg.Add(1)` concurrently with `m.wg.Wait()`. Also, if `run()` exits early on `isTerminalAuthError` (line 492), `m.cancel` is never reset to `nil`.

pkg/googlepolicy/xds_policy_manager.go
```diff
@@ -0,0 +1,1005 @@
...
+	// 2. Activate in memory (googlepolicyprocessor immediately picks this up).
+	SetActivePolicySet(policySet)
+
+	// 3. Send xDS ACK.
+	if err := m.sendACK(stream, node, resp.GetVersionInfo(), resp.GetNonce()); err != nil {
+		m.logger.Error("Failed to send xDS ACK",
+			zap.String("version", resp.GetVersionInfo()),
+			zap.Error(err),
+		)
+	}
```
rafaelwestphal: When a revision contains a mix of valid and invalid policies (`err != nil && len(policySet.Policies) > 0`), this applies the valid subset (`SetActivePolicySet(policySet)`) **and then sends an `ACK` (`error_detail = nil`) instead of a `NACK`**.
In steady state, a fleet almost always has $\ge 1$ baseline policy (`len(policySet.Policies) >= 1`), which means any broken policy added thereafter will be ACKed to the control plane and hidden from the author.
Should we still call `SetActivePolicySet(policySet)` (fail-open) and send `m.sendNACK(...)` whenever `errors.Join(extractErr, makeErr) != nil`?
```suggestion
	// 2. Activate in memory (googlepolicyprocessor immediately picks this up).
	SetActivePolicySet(policySet)
	// 3. Send xDS NACK if any policy failed to decode/load; otherwise ACK.
	if err := errors.Join(extractErr, makeErr); err != nil {
		m.sendNACK(stream, node, resp.GetNonce(), err)
	} else if err := m.sendACK(stream, node, resp.GetVersionInfo(), resp.GetNonce()); err != nil {
		m.logger.Error("Failed to send xDS ACK",
			zap.String("version", resp.GetVersionInfo()),
			zap.Error(err),
		)
	}

rafaelwestphal (changes requested):

rafaelwestphal (approved):
Approved! We can address the NACK/ACK on a follow up

pkg/googlepolicy/policy.go
XuechunHou: Done.

pkg/googlepolicy/xds_policy_manager.go
XuechunHou: Done

pkg/googlepolicy/xds_policy_manager.go
XuechunHou: Will address in a followup PR.

pkg/googlepolicy/registry.go
XuechunHou: You're right that neither map is guarded, and that both are read from the xDS stream goroutine on every revision.

There's no race today, though. All five registrations — logfilter, metricfilter, tracefilter, selfmetrics, gcpdestination — are in init(), and package initialisation completes before main, so every write is ordered before any goroutine that later reads them. The contrast with policySetMu is real but intentional: activePolicySet has two runtime writers (the file and xDS managers) and changes on every revision; these two maps have none after startup.

What was missing is that nothing said so. RegisterPolicyDriver is an exported extension point with no hint about when it's safe to call, and the failure mode if someone registers lazily isn't a quiet degradation — it's fatal error: concurrent map read and map write taking the process down. I've documented the contract in 2e3ede38.

PR #667, opened 2026-09-14
Implement LogFilterPolicy and wire googlepolicy into GBOC

rafaelwestphal:


pkg/googlepolicy/logfilter/policy.go
```diff
@@ -0,0 +1,355 @@
...
+				return nil, false
+			}
+			return matcher.EvalPath(ctx.Resource.Attributes(), steps)
+		}, nil
+
+	case *policyv1alpha1.LogFieldSelector_ScopeAttribute:
+		steps, err := matcher.CompilePath(t.ScopeAttribute, "scope")
+		if err != nil {
+			return nil, err
+		}
```
dashpole, line 255: When a record or scope field is unset (like `SeverityNumber == 0` or empty `SeverityText`), `compileExtractor` returns a non-nil zero value (`int64(0)` or `""`) with `exists = false`, so `!exists && val == nil` evaluates to `false` and causes missing fields to falsely match `<`/`<=` or regex predicates (e.g., dropping logs with unspecified severity when filtering `severity_number < 9`). Should we check `if !exists` instead across these value predicates (lines 250, 262, 274, 287, 300, 313, 326)?
```suggestion
			if !exists {
```

pkg/googlepolicy/logfilter/policy.go
```diff
@@ -0,0 +1,355 @@
...
+		return nil, fmt.Errorf("failed to unmarshal into LogFilterPolicy proto: %w", err)
+	}
+
+	return NewPolicyFromProto(pb)
+}
+
+var (
+	// ErrNilProto is returned when a nil LogFilterPolicy protobuf is provided.
+	ErrNilProto = errors.New("log filter policy proto cannot be nil")
+	// ErrMissingAction is returned when the policy action is unspecified or invalid.
```
dashpole, line 76: Using `DiscardUnknown: false` will reject policy JSON payloads if newer server-side policies add fields that an older collector does not recognize yet. Should we set `DiscardUnknown: true` to preserve forward compatibility?
```suggestion
		DiscardUnknown: true,
```

pkg/googlepolicy/logfilter/policy.go
```diff
@@ -0,0 +1,355 @@
...
+		return nil, ErrMissingAction
+	}
+
+	if len(pb.GetMatches()) == 0 {
+		return nil, ErrMissingMatchers
+	}
+
+	matchers := make([]compiledMatcher, 0, len(pb.GetMatches()))
+	for i, m := range pb.GetMatches() {
+		cm, err := compileMatcher(m)
```
dashpole, line 116: Since `MakePolicySet` keys the active policy map by `p.PolicyName()` (`pb.GetId()`), policies with an empty `id` will collide under `""` and silently overwrite each other. Might be good to reject empty IDs here:
```suggestion
	}
	if pb.GetId() == "" {
		return nil, errors.New("log filter policy id cannot be empty")
	}
```

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
```diff
@@ -210,6 +213,26 @@ func (p *provider) evaluateActivePolicySet(ctx context.Context) (*confmap.Retrie
...
+	}
+	gpCm := confmap.New()
+	if err := gpCm.Marshal(googlePolicyConf); err != nil {
+		return nil, fmt.Errorf("failed to marshal googlepolicy processor config: %w", err)
+	}
+	if err := conf.Merge(cleanConf(gpCm)); err != nil {
+		return nil, fmt.Errorf("failed to merge googlepolicy processor config: %w", err)
+	}
+
+	preProcessLogIDs := append([]component.ID{googlePolicyID}, destPolicy.PreProcessLogIDs()...)
```
dashpole, line 235: Prepending `googlePolicyID` globally applies user transformation policies to `BuiltInSelfMetricsPolicy` (`logs/default_self_metrics` and `metrics/default_self_metrics`), which could cause customer filter policies to accidentally drop collector health logs or self-observability metrics. Should `selfmetrics` bypass `googlepolicy`?

pkg/googlepolicy/logfilter/policy.go
```diff
@@ -0,0 +1,355 @@
...
+	"go.opentelemetry.io/collector/pdata/plog"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+// PolicyType is the registered policy type identifier for log filter policies.
+const PolicyType = "log_filter"
+
+func init() {
+	if err := googlepolicy.RegisterPolicyDriver(PolicyType, &Driver{}); err != nil {
```
dashpole, line 38: Quick nit: `PolicyTypeShort` and `ErrMissingMatches` (line 91) look like redundant aliases for `PolicyType` and `ErrMissingMatchers` in a new package, we can probably remove them and use the canonical names in tests.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
rafaelwestphal: Yeah, i think we don't want policies applied to the self-observability. I was using self-logs to test the policies so i forgot to remove hehe

components/google-built-opentelemetry-collector/processor/googlepolicyprocessor/evaluator.go
```diff
@@ -15,391 +15,63 @@
...
 
 import (
-	"bytes"
 	"fmt"
-	"regexp"
-	"strconv"
-	"strings"
 
-	policyv1alpha1 "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/gen/go/policy/v1alpha1"
 	"github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy"
```
dashpole: Since `logfilter`, `metricfilter`, and `tracefilter` register their drivers in `init()`, we need to blank-import them in non-test code so `googlepolicy.LoadPolicy` can resolve `log_filter`, `metric_filter`, and `trace_filter` policies at runtime:
```suggestion
	"github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy"
	_ "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy/logfilter"
	_ "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy/metricfilter"
	_ "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy/tracefilter"
```

XuechunHou:
Great work refactoring the filter policy evaluation into `pkg/googlepolicy` and wiring `googlepolicyprocessor` into `googlecontrolplaneprovider`!

While reviewing the changes, I noticed a few bugs and edge cases worth addressing:

### 1. Missing blank imports (`_`) for `logfilter`, `metricfilter`, and `tracefilter` in production code
Each filter package registers its driver (`"log_filter"`, `"metric_filter"`, `"trace_filter"`) via `func init()` calling `googlepolicy.RegisterPolicyDriver(...)`.

However, **no non-test `.go` file currently imports `pkg/googlepolicy/logfilter`, `pkg/googlepolicy/metricfilter`, or `pkg/googlepolicy/tracefilter`**:
- `googlepolicyprocessor` was refactored to use the `LogPolicyEvaluator`, `MetricPolicyEvaluator`, and `TracePolicyEvaluator` interfaces in `pkg/googlepolicy`, so its non-test files no longer import the subpackages.
- `googlecontrolplaneprovider/provider.go` imports `gcpdestination` and `selfmetrics` (which registers their drivers), but does not import the three filter packages.

Running `go list -deps` on both `googlecontrolplaneprovider` and `googlepolicyprocessor` confirms none of the three filter packages are compiled into the non-test dependency graph. `go test` passes because `evaluator_test.go` and `processor_test.go` explicitly import them, triggering `init()` in the test binary. In a production build (or when `googlecontrolplaneprovider` loads a filter policy JSON file from disk), `MakePolicySet` -> `LoadPolicy` fails with:
```
no driver found for policy type: log_filter
```
**Suggestion:** Add blank imports in `googlecontrolplaneprovider/provider.go` (and/or `googlepolicyprocessor/factory.go`):
```go
import (
	_ "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy/logfilter"
	_ "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy/metricfilter"
	_ "github.com/GoogleCloudPlatform/opentelemetry-operations-collector/pkg/googlepolicy/tracefilter"
)
```

---

### 2. Untyped metrics (`pmetric.MetricTypeEmpty`) bypass filter evaluation in `TransformMetrics`
In `pkg/googlepolicy/metricfilter/policy.go` (lines 326–336), `compileDescriptorFieldExtractor` explicitly supports matching untyped metrics (`pmetric.MetricTypeEmpty` renders as `"UNSPECIFIED"` with `exists = false`).

However, in `googlepolicyprocessor/metric_transformer.go` (`transformMetricDataPoints`, lines 61–77):
```go
	switch m.Type() {
	case pmetric.MetricTypeGauge:
		datapointCount = m.Gauge().DataPoints().Len()
	// ... Sum, Histogram, ExponentialHistogram, Summary ...
	default:
		return false
	}
```
When `m.Type() == pmetric.MetricTypeEmpty`, the switch hits `default: return false` and immediately returns `false` (KEEP) without evaluating any metric policies.
**Suggestion:** Handle `case pmetric.MetricTypeEmpty:` (leaving `datapointCount = 0`) so execution falls through to `e.evalMetricPolicies(ctx, true)`.

---

### 3. Null attributes (`pcommon.ValueTypeEmpty`) are converted to `""` by `matcher.ValueToAny`
In `pkg/googlepolicy/internal/matcher/matcher.go` (lines 281–305):
```go
func ValueToAny(v pcommon.Value) any {
	switch v.Type() {
	case pcommon.ValueTypeStr:
		return v.Str()
	// ...
	default:
		return v.AsString()
	}
}
```
When an attribute exists in `pcommon.Map` with a null/empty value (`attrs.PutEmpty("key")`, where `v.Type() == pcommon.ValueTypeEmpty`), `ValueToAny` hits `default: return v.AsString()`, returning `""` (a Go `string`). As a result, `EvalPath` returns `("", true)`, causing string predicates like `equals: {string_value: ""}` or `regex: "^$"` to falsely match null attributes.
**Suggestion:** Add `case pcommon.ValueTypeEmpty: return nil` in `ValueToAny`.

---

### 4. Re-evaluating instrument-level metric policies per datapoint
In `googlepolicyprocessor/metric_transformer.go`, when `e.hasDatapointMetricPolicies` is `true`, `pruneDatapoints` calls `evalMetricPolicies(ctx, false)` for every datapoint:
```go
func (e *Evaluator) evalMetricPolicies(ctx MetricContext, instrumentOnly bool) bool {
	var hasDrop bool
	for _, p := range e.metricPolicies {
		if instrumentOnly && p.IsDatapointLevel() {
			continue
		}
		switch p.EvaluateMetric(ctx) { // Evaluates all policies (including instrument-level) per datapoint
```
Because instrument-level policies (`!p.IsDatapointLevel()`) do not inspect `ctx.DatapointAttributes`, their outcome is invariant across all datapoints of a metric. Re-evaluating instrument-level regexes and resource/scope attribute lookups for every datapoint in a metric adds unnecessary overhead on the hot path.
**Suggestion:** Evaluate instrument-level policies once per metric prior to `pruneDatapoints` (if an instrument-level policy returns `EvalDrop` and no datapoint-level policy has `ACTION_KEEP`, the entire metric can also be dropped immediately without iterating datapoints).

---

### 5. Minor doc / comment nits
- **`pkg/googlepolicy/metricfilter/policy.go` (lines 282–287)**: The comment on `compileDescriptorFieldExtractor` states:
  > *"A field that is set to its default therefore reports exists=false while still returning its rendered value, so that string and numeric predicates can match it (see matcher.present)."*

  After commit `769ee1d5`, `Name`, `Description`, and `Unit` return `(nil, false)` when empty rather than `("", false)`, so string predicates (`equals: ""`) no longer match them.
- **`pkg/googlepolicy/registry.go` (lines 58–60)**: The comment on `LoadPolicy` states *"the drivers decode strictly so that field typos fail fast"*, which is outdated now that `DiscardUnknown: true` is used. (Note: one side effect of `DiscardUnknown: true` is that a typo on an optional boolean like `"negated": true` on an `ACTION_DROP` policy will be silently discarded, inverting an allowlist drop rule.)

rafaelwestphal:
I resolved all the opened issues and pushed a new versaion for review

XuechunHou (approved):

PR #668, opened 2026-09-15
pkg/event: Add events package with Weaver schema

braydonk:
This PR adds an event package that all components can use to record events according to a Weaver schema we maintain. The code to record the events is generated directly from the schema. The schema uses Weaver's v2 format.

A skill is included so that agents know how to properly add and use new event types.

Another part of this is that the expected built-in policy fallback that I wanted for destination and self metrics policies didn't exist yet, so I added that in as part of this PR so that when I run a manual end-to-end test using this I can get proper fallbacks (since I'm going to test it with bad destination and self-metrics policies but I still want the events to be recorded with the fallbacks).


components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
```diff
@@ -174,122 +200,195 @@ func (p *provider) evaluateActivePolicySet(ctx context.Context) (*confmap.Retrie
...
-		default:
-			sourceConf, err = srcPolicy.Evaluate(ctx)
-		}
+	// If no custom self_metrics policy was present or if it failed evaluation,
+	// fall back to BuiltInSelfMetricsPolicy.
+	if !appliedSelfMetrics {
+		policyConf, err := p.evaluateSingleSourcePolicy(ctx, collectorID, fleetID, BuiltInSelfMetricsPolicy, destPolicy)
 		if err != nil {
-			return nil, fmt.Errorf("failed to evaluate source policy %q: %w", sp.PolicyName(), err)
+			p.recordPolicyEvaluateError(ctx, BuiltInSelfMetricsPolicy.PolicyName(), activePolicySet.RevisionID, err)
```
dashpole, line 298: If a built-in fallback policy like `BuiltInSelfMetricsPolicy` fails (e.g. `ErrNoFleetID` when `FLEET_ID` is unset), returning an error here causes `evaluateActivePolicySet` to loop and call `RollbackActivePolicySet()` until the entire policy set history is wiped out, while emitting duplicate `policy.evaluate.error` events for each revision ID (and finally an empty `policy.set.revision.id`). Consider handling built-in policy errors without attributing them to `activePolicySet.RevisionID` or triggering rollbacks.

pkg/event/templates/go/events.go.j2
```diff
@@ -0,0 +1,161 @@
...
+package event
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+{% macro go_export_name(key) -%}
+{{ key | pascal_case | replace("Id", "ID") | replace("Url", "URL") | replace("Http", "HTTP") | replace("Grpc", "GRPC") }}
```
dashpole, line 26: Running substring `replace("Id", "ID")` after `pascal_case` or `camel_case` will mangle any word segment starting with `Id` (for example `idle` becomes `IDle` and `identity` becomes `IDentity`). Consider uppercasing initialisms per word segment before joining instead of replacing substrings on the joined name.

pkg/event/templates/go/events.go.j2
```diff
@@ -0,0 +1,161 @@
...
+	fields []zap.Field
+}
+
+// {{ event_base }}EventOption applies optional settings or attributes to a {{ event.name }} event record.
+type {{ event_base }}EventOption func(*{{ config_type }})
+
+// With{{ event_base }}EventContext attaches a context.Context to a {{ event.name }} event record for trace correlation.
+func With{{ event_base }}EventContext(ctx context.Context) {{ event_base }}EventOption {
+	return func(c *{{ config_type }}) {
+		c.fields = append(c.fields, zap.Any("context", ctx))
```
dashpole: If `ctx` is `nil`, `zap.Any("context", nil)` fails the `field.Interface.(context.Context)` check in `otelzap` and gets serialized as a literal `"context": "<nil>"` attribute on the emitted `log.Record`. Maybe guard against a nil context here:
```suggestion
		if ctx != nil {
			c.fields = append(c.fields, zap.Any("context", ctx))
		}
```

pkg/event/templates/go/events.go.j2
```diff
@@ -0,0 +1,161 @@
...
+	err error,
+{%- else %}
+	body string,
+{%- endif %}
+{%- for attr in required_attrs %}
+	{{ go_param_name(attr.key or attr.name) }} {{ go_type(attr) }},
+{%- endfor %}
+	opts ...{{ event_base }}EventOption,
+) {
+	cfg := {{ config_type }}{
```
dashpole: If `logger` is `nil`, calling `logger.Log` below will panic. Adding a quick nil check here makes all generated recording functions safe out of the box so callers don't need defensive nil checks:
```suggestion
	if logger == nil {
		return
	}
	cfg := {{ config_type }}{
```

pkg/event/schema/events.yaml
```diff
@@ -0,0 +1,62 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+file_format: definition/2
+
+attributes:
+  - key: policy.id
```
dashpole, line 18: We should probably put our own custom conventions in the `gcp.` namespace for now (e.g. `gcp.policy.id`, `gcp.policy.set.revision.id`, and `gcp.policy.evaluate.error`) to avoid colliding with any future upstream OpenTelemetry `policy.*` conventions.

pkg/event/schema/events.yaml
```diff
@@ -0,0 +1,62 @@
...
+    requirement_level: recommended
+    brief: Recorded when an error occurs while evaluating a policy.
+    note: |
+      This event is recorded whenever policy evaluation fails.
+      The event body MUST contain the human-readable failure message describing the evaluation error.
+      Timestamp MUST be set to the time when the evaluation failure occurred.
+      Default severity level is ERROR.
+    annotations:
+      severity: error
+    attributes:
```
dashpole, line 42: No need to change in this PR, but in the future we might want to add standard OTel `error.type` (or `gcp.policy.class`) as an optional attribute here so downstream telemetry queries can filter or group evaluation failures without regex-parsing the body string.

pkg/event/templates/go/weaver.yaml
```diff
@@ -0,0 +1,48 @@
...
+    boolean[]: zap.Bools
+    string[]: zap.Strings
+  zap_levels:
+    debug: zapcore.DebugLevel
+    info: zapcore.InfoLevel
+    warn: zapcore.WarnLevel
+    error: zapcore.ErrorLevel
+templates:
+  - template: "events.go.j2"
+    filter: .registry
```
dashpole, line 46: Since `manifest.yaml` defines `schema_url`, `filter: .registry` currently drops it from the template context. If we change this filter to include `.schema_url`, `events.go.j2` can export `const SchemaURL = "{{ ctx.schema_url }}"` so callers initializing `otelzap.NewCore` can attach `otelzap.WithSchemaURL(event.SchemaURL)`:
```suggestion
    filter: "{schema_url: .schema_url, attributes: .registry.attributes, events: .registry.events}"
```

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/provider.go
```diff
@@ -85,6 +86,31 @@ func newProvider(set confmap.ProviderSettings) confmap.Provider {
 	}
 }
 
+func (p *provider) recordPolicyEvaluateError(ctx context.Context, policyID string, revisionID string, err error) {
```
dashpole, line 89: Since `Retrieve()` runs during config resolution before `service.New()` initializes `otelconftelemetry` or starts the `default_self_metrics` receiver, `p.logger` here is only the bootstrap stderr logger. We probably want startup policy error events to go to both stderr (in case we fail to boot) and OTLP (via the fallback pipeline once booted), which may require buffering events recorded during `Retrieve()` and flushing them once the collector telemetry pipeline starts.

PR #674, opened 2026-09-18
Policy collector tests

rafaelwestphal:
## Summary
Adds an end-to-end integration test suite (`integration_test/policy_test/`) for verifying Google-Built OpenTelemetry Collector (`otelcol-google`) policy evaluation across **Logs, Metrics, and Traces** (`log_filter`, `metric_filter`, and `trace_filter` policies).
Each test case launches a real `otelcol-google` subprocess configured via `--config=googlecontrolplane:file:<policies_dir>`, sends OTLP gRPC batches (`input_{logs,metrics,traces}.yaml`) into the collector pipeline, captures exported telemetry in an in-process mock OTLP gRPC server, normalizes timestamps/attribute ordering, and asserts the output against golden files (`expected_{logs,metrics,traces}.yaml`).
Also implements `SelfMetricsPolicy.Validate()` (`p.Name != ""` validation) in `components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/selfmetrics/selfmetrics_policy.go`.

## How It Works
```
testdata/<case>/input_{logs,metrics,traces}.yaml
       │
       ▼ (OTLP gRPC Client)
[ otelcol-google ]
  ├── --config=googlecontrolplane:file:<policies_dir>
  └── pipeline: otlp/test_in ──▶ googlepolicy ──▶ otlp_grpc/test_out
                                                         │
                                                         ▼ (OTLP gRPC)
                                              [ In-Process Mock Server ]
                                                         │
                                                         ▼ (cmp.Diff)
                               testdata/<case>/expected_{logs,metrics,traces}.yaml
```
---
### 1. Build `otelcol-google` (if not already built)
From the repository root on `gboc-experiment-policy`:
```bash
cd google-built-opentelemetry-collector && GOWORK=off make collector-build
```
### 2. Run all integration tests
From `integration_test/policy_test/`:
```bash
cd integration_test/policy_test
GOWORK=off go test -v ./...
```
By default, the harness automatically looks up `../../google-built-opentelemetry-collector/otelcol-google`. You can also point to any custom binary via `--otelcol` or `OTELCOL`:
```bash
GOWORK=off go test -v ./... --otelcol=/path/to/otelcol-google
# or
OTELCOL=/path/to/otelcol-google GOWORK=off go test -v ./...
```
---
## How to Update Golden Files
To regenerate all `expected_{logs,metrics,traces}.yaml` golden files after modifying policies or inputs, pass `-update` (or set `UPDATE_GOLDEN=true`):
```bash
cd integration_test/policy_test
GOWORK=off go test -v ./... -update
# or via Makefile:
make update-golden
```
---
## How to Add a New Test Case
Create a new directory under `integration_test/policy_test/testdata/<case_name>/` containing:
1. `policies/*.json` — one or more policy JSON files loaded by `googlecontrolplane:file:<policies_dir>`.
2. One or more input telemetry files (in OTLP JSON/YAML structure):
   - `input_logs.yaml`
   - `input_metrics.yaml`
   - `input_traces.yaml`
3. Run `GOWORK=off go test -v ./... -update` to generate the corresponding `expected_{logs,metrics,traces}.yaml` golden file(s).
---
## Included Test Cases
| Signal | Test Case | What It Verifies |
| :--- | :--- | :--- |
| **Logs** | `drop_by_body_equals` | Drops log records where `LOG_RECORD_FIELD_BODY == "DROP_THIS_LOG_LINE"` while retaining non-matching logs. |
| **Logs** | `drop_all_logs` | Drops all records matching `resource_attribute` (`service.name`) and asserts an empty exported batch (`[]`). |
| **Logs** | `keep_exemption_precedence` | Verifies `ACTION_KEEP` exemption (`env == "prod"`) takes precedence over a broader `ACTION_DROP` (`severity_number < 9`). |
| **Logs** | `nested_attribute_and_regex` | Traverses nested map attributes (`path: ["http", "headers", "authorization"]`) with regex matching (`^Bearer secret-.*`). |
| **Metrics** | `metric_drop_by_descriptor_name` | Evaluates instrument-level `METRIC_DESCRIPTOR_FIELD_NAME` regex (`^internal\.debug\..*`) to drop matching metrics while keeping others. |
| **Metrics** | `metric_datapoint_attribute_and_keep_exemption` | Prunes individual datapoints by `datapoint_attribute` (`env == "ephemeral"`) while exempting datapoints matching `ACTION_KEEP` (`service.tier == "critical"`). |
| **Traces** | `trace_drop_healthcheck_spans` | Drops spans matching compound `SPAN_RECORD_FIELD_KIND == "SERVER"` AND `span_attribute` `http.target` regex `^/(healthz\|readyz)$`. |
| **Traces** | `trace_keep_error_and_root_span_precedence` | Verifies `ACTION_KEEP` exemptions on both error spans (`SPAN_RECORD_FIELD_STATUS_CODE == "ERROR"`) and root spans (`SPAN_RECORD_FIELD_PARENT_SPAN_ID == ""`) over `ACTION_DROP` (`SPAN_RECORD_FIELD_KIND == "INTERNAL"`).


components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/selfmetrics/selfmetrics_policy.go
```diff
@@ -70,7 +70,10 @@ func (p *SelfMetricsPolicy) PolicyClass() googlepolicy.PolicyClass {
 }
 
 func (p *SelfMetricsPolicy) Validate() error {
-	panic("unimplemented")
+	if p.Name == "" {
+		return errors.New("policy must be named")
+	}
+	return nil
```
dashpole, line 87: Might be worth also validating `p.Port` here so an omitted (`0`) or out-of-range port fails validation early instead of generating a `localhost:0` receiver and exporter config.
```suggestion
	if p.Port <= 0 || p.Port > 65535 {
		return fmt.Errorf("invalid port %d: must be between 1 and 65535", p.Port)
	}
	return nil
```

integration_test/policy_test/Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: test update-golden
```
dashpole: Since `make workspace` adds `integration_test/policy_test` to `go.work`, root targets like `make update-go-module-in-all` will run `make -C integration_test/policy_test update-go-module` and `tidy`. Consider including `../../make/maintenance.mk` and adding a `tidy` target so those repo-wide targets don't fail:
```suggestion
include ../../make/maintenance.mk

.PHONY: test tidy update-golden

tidy:
	go mod tidy
```

integration_test/policy_test/policy_test.go
```diff
@@ -0,0 +1,623 @@
...
+	// Allow in-flight OTLP batches to reach mock exporter before triggering graceful drain.
+	time.Sleep(250 * time.Millisecond)
+
+	if err := cmd.Process.Signal(syscall.SIGTERM); err != nil {
+		t.Logf("warning: failed to send SIGTERM to collector: %v", err)
+	}
+
+	waitDone := make(chan error, 1)
+	go func() {
+		waitDone <- cmd.Wait()
```
dashpole: Calling `cmd.Wait()` before the `stderrPipe` scanner goroutine exits can close `stderrPipe` mid-read and drop trailing log lines. Consider closing a done channel when the scanner goroutine finishes and waiting on it before calling `cmd.Wait()`.

integration_test/policy_test/Makefile
```diff
@@ -0,0 +1,9 @@
...
+
+OTELCOL ?= $(abspath ../../google-built-opentelemetry-collector/otelcol-google)
+
+test:
+	GOWORK=off go test -v ./... -otelcol=$(OTELCOL)
```
dashpole: We should make sure these tests run in CI (right now `make test-all` excludes `integration_test/*` and `presubmit.yaml` doesn't run this suite). Also, if building the full `otelcol-google` binary in presubmit is too heavy, have you considered running the collector in-process with `otelcol.NewCollector` so `go test ./...` can run it directly?

integration_test/policy_test/policy_test.go
```diff
@@ -0,0 +1,623 @@
...
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+		caseName := entry.Name()
+		t.Run(caseName, func(t *testing.T) {
+			t.Parallel()
+			runPolicyTestCase(t, otelcolBin, caseName)
```
dashpole: It would also be great to add a test case (here or as a follow-up) that updates a policy file in `policiesDir` while the collector is running to verify `fsnotify` hot-reloading end-to-end.

integration_test/policy_test/policy_test.go
```diff
@@ -0,0 +1,623 @@
...
+		if _, err := client.Export(ctx, ptraceotlp.NewExportRequestFromTraces(inputTraces)); err != nil {
+			stderrMu.Lock()
+			logs := stderrLog.String()
+			stderrMu.Unlock()
+			t.Fatalf("failed to export traces to collector: %v\nstderr:\n%s", err, logs)
+		}
+	}
+
+	// Allow in-flight OTLP batches to reach mock exporter before triggering graceful drain.
+	time.Sleep(250 * time.Millisecond)
```
dashpole: If we configure `sending_queue: {enabled: false}` on `otlp_grpc/test_out` in `test_pipeline.yaml`, `client.Export()` will flush synchronously to `mockSrv` before returning, which could let us avoid sleeping 250ms here.

components/google-built-opentelemetry-collector/provider/googlecontrolplaneprovider/policies/selfmetrics/selfmetrics_policy.go
rafaelwestphal: Fixed

integration_test/policy_test/policy_test.go
```diff
@@ -0,0 +1,1007 @@
...
+
+	factories.Receivers, err = otelcol.MakeFactoryMap[receiver.Factory](
+		otlpreceiver.NewFactory(),
+	)
+	if err != nil {
+		return otelcol.Factories{}, err
+	}
+	factories.ReceiverModules = makeModulesMap(factories.Receivers, map[component.Type]string{
+		otlpreceiver.NewFactory().Type(): "go.opentelemetry.io/collector/receiver/otlpreceiver v0.160.0",
+	})
```
LujieDuan: The `Modules` map might be ok to omit for this in process collector. The factory map should already provide the collector the right info to start each component. 

Would the collector run if we do not provide the ReceiverModules/ExporterModules/ProcessorModules/ProviderModules?

rafaelwestphal: Yes, it runs without them! Those *Modules maps are only used by the otelcol components CLI subcommand to print module version metadata, and col.Run(ctx) only needs the factory maps. Removed ReceiverModules, ExporterModules, ProcessorModules, ExtensionModules, ProviderModules, and makeModulesMap in 3b021ddd

LujieDuan (approved):
LGTM

